PR #8, opened 2022-06-01
Move vite to dependencies

ankurk91:
Hi @jessarcher 

Lets move `vite` package to `dependencies`

* Force end users to use maintainer's defined version of `vite` :muscle:
* Just one package to install by end users :pinching_hand:

Laravel-Mix does not ask you to install `webpack` separately, right? :smirk:

#### Additional things to be done
* Specify the minimum node.js version required in docs
* Specify the minimum Laravel Framework version to use `@vite` blade directive

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -9,7 +9,7 @@
 First you will need to install [Vite](https://vitejs.dev/) and the [Laravel Vite Plugin](https://www.npmjs.com/package/laravel-vite-plugin) using your npm package manager of choice:
 
 ```shell
-npm install --save-dev vite laravel-vite-plugin
+npm install --save-dev laravel-vite-plugin
 ```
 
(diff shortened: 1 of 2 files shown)


emargareten:
also remove `plugin` from the package name

ankurk91:
@emargareten 
I think the package name `laravel-vite` is already acquired

https://www.npmjs.com/package/laravel-vite

jessarcher:
Hi @ankurk91,

Mix doesn't require you to install webpack because Mix is a wrapper around webpack.

This package is a plugin for Vite, so users are interacting with Vite directly. It's the same process for installing the Vue or React plugins for Vite.

timacdonald:
@ankurk91 FYI the plugin does have a peer dependency on Vite.

ankurk91:
Thanks for clarification

PR #10, opened 2022-06-09
Update test case

webfansplz:
The test case for prefixes the `base` with `ASSET_URL` only work in the `production` mode.


taylorotwell:
Marking as draft until @jessarcher reviews.

jessarcher (approved):
Thanks :+1:

PR #11, opened 2022-06-09
Update resolvePageComponent helper to work with both glob and glogEager

Esirei:
This pull request updates the `resolvePageComponent` inertia helper to work with the object returned both by the glob and globEager imports.

### DIFF
--- a/src/inertia-helpers/index.ts
+++ b/src/inertia-helpers/index.ts
@@ -1,4 +1,4 @@
-export async function resolvePageComponent<T>(path: string, pages: Record<string, T | () => Promise<T>>) {
+export async function resolvePageComponent<T>(path: string, pages: Record<string, T | (() => Promise<T>)>) {
     const page = pages[path]
     if (typeof page === 'undefined') {
         throw new Error(`Page not found: ${path}`)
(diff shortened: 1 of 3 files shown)


jessarcher:
Hi @Esirei, thanks for the PR!

Could you please explain the use case and benefit of this?

Esirei:
> Hi @Esirei, thanks for the PR!
> 
> Could you please explain the use case and benefit of this?

Hello @jessarcher , unlike the glob import which compiles each page/component/js into separate files and requests them on demand (lazy loaded), the globEager import compiles all into a single app.js file. Of course this will have a slower initial website load, but faster page responses, it depends on the developer which tradeoff they want.

This PR updates the helper function to work with either technique, I just need to figure out how to tell typescript that the generic cannot be a function to fix the failed tests.

jessarcher:
Thanks, @Esirei. That makes sense :+1:

Esirei:
Hi @jessarcher, I finally had some time to get the issue out of the way, also added tests for the helper.

timacdonald (approved):

jessarcher:
Thanks, @Esirei!

PR #13, opened 2022-06-13
Fix manifest handling in SSR mode

jessarcher:
The manifest augmentation added in #9 broke SSR builds because the manifest file isn't typically created in this mode.

This PR addresses this by only augmenting the manifest when the manifest is enabled for the build.

It will also now use the manifest name from the resolved config if it exists.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -170,7 +170,13 @@ export default function laravel(config?: string|string[]|Partial<PluginConfig>):
             return null
         },
         writeBundle() {
-            const manifestPath = path.resolve(resolvedConfig.root, resolvedConfig.build.outDir, 'manifest.json')
+            const manifestConfig = resolveManifestConfig(resolvedConfig)
+
+            if (manifestConfig === false) {
+                return;
+            }
+
+            const manifestPath = path.resolve(resolvedConfig.root, resolvedConfig.build.outDir, manifestConfig)
             const manifest = JSON.parse(fs.readFileSync(manifestPath).toString())
             const newManifest = {
                 ...manifest,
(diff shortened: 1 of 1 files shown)


src/index.ts
```diff
@@ -170,7 +170,12 @@ export default function laravel(config?: string|string[]|Partial<PluginConfig>):
             return null
         },
         writeBundle() {
-            const manifestPath = path.resolve(resolvedConfig.root, resolvedConfig.build.outDir, 'manifest.json')
+            const manifestConfig = resolvedConfig.build.ssr ? resolvedConfig.build.ssrManifest : resolvedConfig.build.manifest;
+            if (manifestConfig === false) {
+                return;
+            }
+            const manifestFile = typeof manifestConfig === 'string' ? manifestConfig : 'manifest.json'
```
timacdonald, line 177: Just wanna confirm that the default manifest for SSR is also `'manifest.json'`, as the docs don't specify.

timacdonald: Looks like the default name in SSR is `ssr-manifest.json`

jessarcher: Good catch - it should be `ssr-manifest.json`.

https://github.com/vitejs/vite/blob/0e20949dbf0ba38bdaefbf32a36764fe29858e20/packages/vite/src/node/ssr/ssrManifestPlugin.ts#L82

timacdonald:
casually force pushing to your branch 😬

src/index.ts
```diff
@@ -259,3 +265,20 @@ function resolveOutDir(config: PluginConfig, ssr: boolean): string|undefined {
 
     return path.join(config.publicDirectory, config.buildDirectory)
 }
+
+function resolveManifestConfig(config: ResolvedConfig): string|false
```
timacdonald, line 269: This return type is dedicated to @michaeldyrynda ❤️

jessarcher:
Changes look good!

timacdonald (approved):
@jessarcher I've made some tweaks to accommodate the differing manifest filename in ssr mode, when just passing `true`.

Would love you to re-review before we merging.

timacdonald:
haha. beat me to it.

PR #14, opened 2022-06-13
Remove default entry point

jessarcher:
To avoid a flash of unstyled content with Blade apps, we needed to specify a CSS entry point so that we can load a `<link rel="stylesheet">` tag in HMR mode.

However, for SPAs, this results in a different flash of unstyled content due to the DOM being updated slightly before the new `<link>` tag is loaded. Therefore it is preferable with SPAs to import the CSS via the JavaScript entry point as we had previously.

Given that it doesn't seem possible to create an optimal experience for both SPAs and Blade apps with the same default entry point(s), this PR removes the default entirely so that they must be specified explicitly.

This will of course be configured appropriatly for the stack by Breeze and Jetstream.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -34,7 +34,10 @@ import laravel from 'laravel-vite-plugin'
 
 export default defineConfig({
     plugins: [
-        laravel(),
+        laravel([
+            'resources/css/app.css',
+            'resources/js/app.js',
+        ]),
         // react(),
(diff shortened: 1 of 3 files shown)


timacdonald (approved):
I still feel it would be nice to provide defaults for those that we can and just make blade apps specify entry points, but LGTM and we can circle back if we feel it is worthwhile.

Issue #15, opened 2022-06-13
Vite dev server exits without message (fix in report)


ericwang401:
- Laravel Vite Plugin Version: 0.1.3
- Laravel Version: 9.17.0
- Node Version: 18.3.0
- NPM Version: 8.11.0

### Description:

When running the Vite dev server using `npm run dev`, the development server starts up with
```
root@2731f07e003e:/app# npm run dev

> dev
> vite --host 0.0.0.0


  vite v2.9.12 dev server running at:

  > Local:    http://localhost:3000/
  > Network:  http://172.21.0.4:3000/

  ready in 506ms.


  Laravel v9.17.0 

  > APP_URL: http://localhost
```
But closes unexpectedly a second after running the command with no messages or error messages whatsoever.

I did some debugging and commented out `process.exit();` in
```
const clean = () => {
    if (fs_1.default.existsSync(hotFile)) {
        fs_1.default.rmSync(hotFile);
    }
    process.exit();
};
```
and this finally made the dev server spit out the following error
```
root@2731f07e003e:/app# npm run dev

> dev
> vite --host 0.0.0.0


  vite v2.9.12 dev server running at:

  > Local:    http://localhost:3000/
  > Network:  http://172.21.0.4:3000/

  ready in 448ms.


  Laravel v9.17.0 

  > APP_URL: http://localhost
node:internal/errors:466
    ErrorCaptureStackTrace(err);
    ^

Error: ENOSPC: System limit for number of file watchers reached, watch '/app/vendor/symfony/mailer/Transport/Smtp/Auth/XOAuth2Authenticator.php'
    at FSWatcher.<computed> (node:internal/fs/watchers:244:19)
    at Object.watch (node:fs:2306:34)
    at createFsWatchInstance (/app/node_modules/vite/dist/node/chunks/dep-8f5c9290.js:45512:17)
    at setFsWatchListener (/app/node_modules/vite/dist/node/chunks/dep-8f5c9290.js:45559:15)
    at NodeFsHandler$1._watchWithNodeFs (/app/node_modules/vite/dist/node/chunks/dep-8f5c9290.js:45714:14)
    at NodeFsHandler$1._handleFile (/app/node_modules/vite/dist/node/chunks/dep-8f5c9290.js:45778:23)
    at NodeFsHandler$1._addToNodeFs (/app/node_modules/vite/dist/node/chunks/dep-8f5c9290.js:46020:21)
Emitted 'error' event on FSWatcher instance at:
[...]

PR #16, opened 2022-06-13
Add docs for dev server silently exiting

ericwang401:
This pull request addresses https://github.com/laravel/vite-plugin/issues/15 

This request proposes a workaround for the dev server silently exiting.

It can save countless developers' hair from being forcefully removed haha.


jessarcher:
If we merge #17 then the developer will get a helpful error message that they can search, so I'm not we would still need this PR as the issue won't have anything to do with this plugin.

PR #18, opened 2022-06-15
Prevent SSR build from externalizing Inertia helpers

jessarcher:
This PR fixes an issue where Vite SSR builds would fail to run due to the ESM Inertia helpers being incorrectly detected as externalizable (I know, right...)

The [`ssr` options](https://vitejs.dev/config/#ssr-options) are experimental/alpha so they aren't included in the type definition, hence the ts-ignores.

https://vitejs.dev/guide/ssr.html#ssr-externals

Tested with Breeze and Jetstream.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -2,7 +2,7 @@ import fs from 'fs'
 import { AddressInfo } from 'net'
 import path from 'path'
 import colors from 'picocolors'
-import { Plugin, loadEnv, UserConfig, ConfigEnv, Manifest, ResolvedConfig } from 'vite'
+import { Plugin, loadEnv, UserConfig, ConfigEnv, Manifest, ResolvedConfig, SSROptions } from 'vite'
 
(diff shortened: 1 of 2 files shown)


src/index.ts
```diff
@@ -293,3 +299,28 @@ function resolveManifestConfig(config: ResolvedConfig): string|false
...
 }
+
+/**
+ * Add the Interia helpers to the list of SSR dependencies that aren't externalized.
+ *
+ * @see https://vitejs.dev/guide/ssr.html#ssr-externals
+ */
+function noExternalInertiaHelpers(config: UserConfig): true|Array<string|RegExp> {
+    /* eslint-disable-next-line @typescript-eslint/ban-ts-comment */
+    /* @ts-ignore */
```
timacdonald, line 310: ya just love to see it

timacdonald (approved):

PR #20, opened 2022-06-17
Fix manifest CSS paths on Windows

### LINKED ISSUE #19: [Windows]: Possible entrypoint path mismatch between the manisfest.json and Illuminate/Foundation/Vite.php
- Laravel Vite Plugin Version: **v0.2.1** _(latest)_
- Laravel Version: **v9.17.0** _(after a `composer create-project laravel/laravel:dev-vite`)_
- Node Version: **v16.13.1**
- NPM Version: **v8.3.0**

### Description:
An [`Unable to locate file in Vite manifest: {$entrypoint}`](https://github.com/laravel/framework/blob/013ec2768a42c92af9a6946ff8bdd19a427b4968/src/Illuminate/Foundation/Vite.php#L54) exception may be thrown on Windows caused by a mismatch between the entrypoint path strings stored in `manisfest.json` and an entry point given to the `@vite` Blade directive.

#### Example
Consider the following bare bones Vite config file:

```ts
export default defineConfig({
    plugins: [
        laravel([
            'resources/css/app.css',
            'resources/ts/app.ts',
        ]),
    ],
});
```

And the `@vite` blade directive usage:

```blade
@vite(['resources/css/app.css', 'resources/ts/app.ts'])
```

This may produce the aforementioned exception.

This is because entrypoint paths in the `manifest.json` file may follow Windows' backslash path separator: 

```jsonc
{
    "resources/ts/app.ts": {
        "file": "assets/app.e6f6eb1a.js",
        "src": "resources/ts/app.ts",
        "isEntry": true,
    },
    "resources\\css\\app.css": { // <-- backslashes here
        "file": "assets/app.6d9d1460.css",
        "src": "resources\\css\\app.css",
        "isEntry": true
    }
}
```

But as here the CSS path given to the directive _(`'resources/css/app.css'`)_ uses the regular slash path separator this causes `$manifest[$entrypoint]` to fail in line 53 of `/laravel/framework/blob/vite/src/Illuminate/Foundation/Vite.php`.

#### Note
This only seem to happen for CSS entry paths.

#### Possible solutions
From what I understand taking a quick look at the source code of each library, the problem could be solved in [line 166](https://github.com/laravel/vite-plugin/blob/c41269239bf8c2c890e63f0d93c60c28e5d9f177/src/index.ts#L166) by doing something similar to:

```ts
const relativeChunkPath = path.relative(resolvedConfig.root, chunk.facadeModuleId).replaceAll('\\', '/');
```

And on the [PHP side](https://github.com/laravel/framework/blob/013ec2768a42c92af9a6946ff8bdd19a427b4968/src/Illuminate/Foundation/Vite.php#L52):

```php
// ...

$tags = collect();

foreach ($entrypoints as $entrypoint) {
    $entrypoint = str_replace('\\', '/', $entrypoint);

    if (!isset($manifest[$entrypoint])) {
        throw new Exception("Unable to locate file in Vite manifest: {$entrypoint}.");
    }

    // ...
}
```

Or only on the PHP side:

```php
// ...

if (!isset($manifests[$manifestPath])) {
    // ...

    $manifests[$manifestPath] = collect(json_decode(file_get_contents($manifestPath), true))->flatMap(
        fn(array $value, string $key): array => [str_replace('\\', '/', $key) => $value]
    );
}

// ...

foreach ($entrypoints as $entrypoint) {
    $entrypoint = str_replace('\\', '/', $entrypoint);

    // ...
}
```

### Question:
Which approach would be better suited to be PRed ?

### Steps To Reproduce:
- Have a Windows machine
- Check out the example scenario given above

jessarcher:
This PR fixes an issue where the plugin would output Windows paths to the manifest for CSS files on Windows systems. These paths should always be POSIX style.

See https://vitejs.dev/guide/api-plugin.html#path-normalization

Fixes #19

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -2,7 +2,7 @@ import fs from 'fs'
 import { AddressInfo } from 'net'
 import path from 'path'
 import colors from 'picocolors'
-import { Plugin, loadEnv, UserConfig, ConfigEnv, Manifest, ResolvedConfig, SSROptions } from 'vite'
+import { Plugin, loadEnv, UserConfig, ConfigEnv, Manifest, ResolvedConfig, SSROptions, normalizePath } from 'vite'
 
 interface PluginConfig {
     /**
@@ -163,7 +163,7 @@ export default function laravel(config: string|string[]|PluginConfig): LaravelPl
                 return null
             }
 
-            const relativeChunkPath = path.relative(resolvedConfig.root, chunk.facadeModuleId)
+            const relativeChunkPath = normalizePath(path.relative(resolvedConfig.root, chunk.facadeModuleId))
 
             cssManifest[relativeChunkPath] = {
                 /* eslint-disable-next-line @typescript-eslint/ban-ts-comment */


timacdonald (approved):
1+1=a...

PR #21, opened 2022-06-17
remove ziggy alias

timacdonald:
This removes the ziggy alias and pushes the concern back to userland.


jessarcher:
Just for future reference - the reason we opted to remove this is if you're using Vue then you'll likely want a different import. See https://github.com/laravel/jetstream/pull/1069/files

PR #22, opened 2022-06-21
Use Vite's envDir option if specified

hailwood:
Vite allows you to pass through a custom option for the envDir.
This is useful in cases where you might have multiple frontend packages and so your vite.config.js is not in the root directory.

This update checks for that config option and if so will use it over the current directory when resolving the APP_URL and laravel version.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -124,10 +124,11 @@ export default function laravel(config: string|string[]|PluginConfig): LaravelPl
                     viteDevServerUrl = `${protocol}://${host}:${address.port}`
                     fs.writeFileSync(hotFile, viteDevServerUrl)
 
-                    const appUrl = loadEnv('', process.cwd(), 'APP_URL').APP_URL
+                    const envDir = resolvedConfig.envDir || process.cwd()
+                    const appUrl = loadEnv('', envDir, 'APP_URL').APP_URL
 
                     setTimeout(() => {
-                        server.config.logger.info(colors.red(`\n  Laravel ${laravelVersion()} `))
+                        server.config.logger.info(colors.red(`\n  Laravel ${laravelVersion(envDir)} `))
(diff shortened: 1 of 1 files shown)


jessarcher:
Hi @hailwood,

I agree that it makes sense to use `envDir` when looking for `.env` files, but I'm not sure that it should also be used to locate `composer.json`.

It's possible in Laravel to specify a different directory for `.env` files, so there's no guarantee that the `envDir` and the Laravel root directory are the same.

The loading of `composer.json` is already in a try/catch, so the worst-case scenario should be that no version information is displayed.

Can we leave the `laravelVersion()` as-is, and just use `envDir` to load the `APP_URL` env var?

hailwood:
Hi @jessarcher,
You are correct of course that you can change where the .env files are in laravel, but since they're both edge cases I wonder what's the most common -
1. vite.config.js file not in root
2. .env file not located next to the composer.json

Happy to revert the laravelVersion change if you believe edge case 2 is more common than edge case 1.
Let me know :)

jessarcher:
Thanks, @hailwood. I agree that the first edge case seems more common, but haven't personally run into either.

It's mostly that I'm just not a fan of using `envDir` to find `composer.json`. We could add a config option to the plugin for the "laravel root", but it doesn't seem worthwhile just to display the version number. The plugin could also recursively look up the directory tree until it finds a `composer.json`, but again that seems like overkill to grab a version number. I think I'd almost rather just not display the Laravel version number, or display the plugin version number instead.

If you wouldn't mind reverting the `laravelVersion()` change I'll get this merged in, and perhaps we can come up with another solution for the Laravel version later.

jessarcher:
Thanks, @hailwood!

hailwood:
No worries :)
Thanks for all your work on this!

PR #25, opened 2022-06-22
Respect server.host variable in hot file

### LINKED ISSUE #23: url stored in hot file does not follow vite conventions
- Laravel Vite Plugin Version: 0.2.2
- Laravel Version: 9.17.0
- Node Version: 18.0.0
- NPM Version: 8.6.0

### Description:
Right now we store the server IP in the hot file, either the IPV4 or the IPV6 variation.
This causes issues when using a self signed localhost valid certificate as we're telling the server to always load from 127.0.0.1.

Vite has [a resolveHostname method we should probably be using](https://github.com/vitejs/vite/blob/v2.9.12/packages/vite/src/node/utils.ts#L711). This will also ensure our loaded url matched the `dev server running at` output from vite.

### Steps To Reproduce:
1. Change the hostname via either `--host localhost` or `server.host`
2. Run `npm run hot`
3. Attempt to load a page using vite
4. Notice that it's requesting `127.0.0.1` instead

jessarcher:
This PR adds support for Vite's `server.host` setting when writing the address to the hot file.

Fixes #23

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -120,7 +120,7 @@ export default function laravel(config: string|string[]|PluginConfig): LaravelPl
                 const isAddressInfo = (x: string|AddressInfo|null|undefined): x is AddressInfo => typeof x === 'object'
                 if (isAddressInfo(address)) {
                     const protocol = server.config.server.https ? 'https' : 'http'
-                    const configHost = typeof resolvedConfig.server.host === 'string' ? resolvedConfig.server.host : false
+                    const configHost = typeof server.config.server.host === 'string' ? server.config.server.host : false
                     const serverAddress = address.family === 'IPv6' ? `[${address.address}]` : address.address
                     const host = configHost || serverAddress
                     viteDevServerUrl = `${protocol}://${host}:${address.port}`
--- a/src/index.ts
+++ b/src/index.ts
@@ -120,7 +120,9 @@ export default function laravel(config: string|string[]|PluginConfig): LaravelPl
                 const isAddressInfo = (x: string|AddressInfo|null|undefined): x is AddressInfo => typeof x === 'object'
                 if (isAddressInfo(address)) {
                     const protocol = server.config.server.https ? 'https' : 'http'
-                    const host = address.family === 'IPv6' ? `[${address.address}]` : address.address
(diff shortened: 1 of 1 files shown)


hailwood:
@jessarcher This resolves the issue if we've manually specified the host,
If we haven't though, we still end up with 127.0.0.1 while vite defaults to localhost.

Perhaps this small update would help?

```js
const host = configHost || (serverAddress === '127.0.0.1' ? 'localhost' : serverAddress)
```

As an aside, I think we should also be setting `server.hmr.host` to the same value, otherwise hmr attempts to use whatever domain we loaded the site under in our browser (e.g. I access the site in dev at https://laravel.lndo.site so it's requesting `https://laravel.lndo.site/__vite_ping and attempting to open a websocket connection to wss://laravel.lndo.site:3000)

jessarcher:
Ahh, I see.

I worry about specifying `localhost` manually as it seems like that has caused issues in the past with IPv4 and IPv6, where on some systems `localhost` will resolve to `::1`, but in your code snippet Vite will only be listening on `127.0.0.1`. See [here](https://github.com/vitejs/vite/blob/8ef733368fd6618a252e44616f7577f593fd4fbc/packages/vite/src/node/utils.ts#L781-L787) for Vite's solution. Unless we implement a similar DNS resolution, it seems safer to use the IP address unless explicitly configured to be a hostname.

In terms of the `__vite_ping` thing, I don't think setting `server.hmr.host` will fix that as Vite 2 doesn't specify any hostname when making that request, so it will always go to the web server rather than the Vite server. This has been fixed in https://github.com/vitejs/vite/pull/6819 but that only made it to Vite 3.

hailwood:
Okay, I don't mind manually specifying localhost in my config, as long as the hot file picks that up (which is what the PR is currently doing) so good to go there.

Re the __vite_ping thing, I wonder if that's more a side effect I end up seeing if the websocket connection fails, which is why I don't see it when setting the hmr host? 
Setting the hmr host definitely corrects the issue for the websocket connection heading to the wrong location though.
Again I don't mind manually specifying that in my config if you'd rather wait and see if anyone else reports it as an issue though :)

PR #26, opened 2022-06-22
Update UPGRADE.md

driesvints:
- Moved commands to install Vue and React plugins to separate code blocks for easy copy/paste using native GitHub buttons
- Added a note about removing the mix build manifest and the rule from the gitignore file
- Added a note to build Vite one time at the end of the migration (should we add some info here when people run into issues like I did?)

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -184,6 +184,12 @@ And you may remove your Mix configuration file:
 rm webpack.mix.js
 ```
 
+To wrap up, you may remove your Mix build manifest as well as the rule from the `.gitignore` config file:
+
+```shell
+rm public/mix-manifest.json
+```
+
 If you are using StyleCI and have ignored the `webpack.mix.js` file in your configuration, you may also like to remove the ignore rule.
 
 ### Optional: Configure Tailwind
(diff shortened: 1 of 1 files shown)


UPGRADE.md
```diff
@@ -184,6 +184,12 @@ And you may remove your Mix configuration file:
 rm webpack.mix.js
 ```
 
+To wrap up, you may remove your Mix build manifest as well as the rule from the `.gitignore` config file:
```
jessarcher, line 187: The Mix manifest isn't included in the default `.gitignore`. I know it's very common though, in addition to ignoring `public/css`, `public/js`, etc. so if we mention this maybe we should also mention removing the old build files and ignores?

UPGRADE.md
```diff
@@ -263,6 +269,16 @@ You may start the SSR server using `node`:
 node storage/ssr/ssr.js
 ```
 
+### Wrapping Up
+
+You should now be able to build your assets using the following command. This will also invoke the Vite server and Vite will watch for file changes:
+
+```shell
+npm run dev
+```
```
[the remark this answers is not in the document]
driesvints: Done!

UPGRADE.md
jessarcher: With Vite, the build assets and manifest are placed in `public/build` rather than `public/css`, `public/js`, etc.

I've added `public/build` to the `.gitignore` for `laravel/laravel` so compiled assets will now be ignored by default.

driesvints: Right! I ended up removing this. People will figure this out on their own 👍

jessarcher (approved):

PR #31, opened 2022-06-28
Update UPGRADE.md

ekrist1:
The the @ alias no longer works after changing from Webpack to Vite. Added instruction how to add resolve alias to Vite.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -170,6 +170,41 @@ See [this tweet](https://twitter.com/youyuxi/status/1362050255009816577) from Vi
 + import Button from './Button.vue'
 ```
 
+### Import Alias in Vite
+
+To make the @ alias work, add a resolve alias to your vite.config.js file
+
+```js
+import { defineConfig } from 'vite'
+import laravel from 'laravel-vite-plugin'
++ const path = require('path');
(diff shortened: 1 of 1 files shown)


jessarcher:
The plugin is supposed to add this common alias for you: https://github.com/laravel/vite-plugin/blob/c1bac96d90c3bcd597761be15b3cfaede46e57d1/src/index.ts#L57-L59

It sounds like this isn't working for you, so rather than document a workaround, I'd like to get this fixed in the plugin for you.

Are you on Windows by any chance? I'm wondering whether we just need to correct the path separators.

jessarcher:
We've tested this with Windows and the default alias provided by the Vite plugin seem to work fine. I'm assuming the issue was with migrating aliases from your `webpack.mix.js` file over to your `vite.config.js`, where you now need to make sure you have a leading `/` on the path. Alternatively, you can remove the alias entirely if you're using the default that the plugin provides.

A fix for the official docs has already been submitted at https://github.com/laravel/docs/pull/8013, and we have added a section to the migration guide at https://github.com/laravel/vite-plugin/pull/35, so I'm going to close this PR.

Thanks for pointing this issue out. Feel free to re-open it if this doesn't solve it for you.

PR #34, opened 2022-06-28
[0.x] Ensure manifest always contains a file key for CSS sources

### LINKED ISSUE #33: Building an empty CSS file generates JS file and bugged manifest
- Laravel Vite Plugin Version: 0.2.1
- Laravel Version: 9.19.0
- Node Version: 16.15.1
- NPM Version: 8.11.0

### Description:

When building (`npm run build`) an empty CSS file, it'll generate a `.js` file in it's place, with a manifest that doesn't make sense and throws an error in the application.

Here's my generated manifest:

```json
{
  "resources/css/app.css": {
    "src": "resources/css/app.css",
    "isEntry": true
  },
  "resources/js/app.js": {
    "file": "assets/app.f40b63e3.js",
    "src": "resources/js/app.js",
    "isEntry": true
  }
}
```

And here's the [StackTrace](https://flareapp.io/share/xPQkaJ25#F65).

I am unsure if this is a Vite specific thing. I can obviously leave out the CSS file in the `@vite` directive since an empty CSS file doesn't make sense. But I still believe this is wrong behavior.

### Steps To Reproduce:

- Generate new Laravel app
- Run run build on an empty `.css` file
- Use the `@vite` directive to pull the bundled CSS file (which has become a JS file)

timacdonald:
This PR fixes an issue when a CSS entry point is an empty file.

When this is the case, the manifest is missing the `file` key, which is used to generate the HTML script / link tags.

This fix ensures that there is always a file key present.

Fixes #33

This brings our Vite 2 workaround in alignment with how Vite 3 handles manifest generation for empty CSS files.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -169,7 +169,7 @@ export default function laravel(config: string|string[]|PluginConfig): LaravelPl
             cssManifest[relativeChunkPath] = {
                 /* eslint-disable-next-line @typescript-eslint/ban-ts-comment */
                 /* @ts-ignore */
-                file: Array.from(chunk.viteMetadata.importedCss)[0],
+                file: Array.from(chunk.viteMetadata.importedCss)[0] ?? chunk.fileName,
                 src: relativeChunkPath,
                 isEntry: true,
             }


jessarcher (approved):

w3roman:
Generates a .js file.

PR #52, opened 2022-07-01
Add a link to the Vite page

DannyFeliz:
A small reference to Vite in case someone is not very familiar and wants to go directly to the official website for more details.

If you think this is not relevant feel free to close this PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@
 
 [Vite](https://vitejs.dev) is a modern frontend build tool that provides an extremely fast development environment and bundles your code for production.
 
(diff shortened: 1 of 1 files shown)


jessarcher:
Good call. Would you mind just removing the second link as we typically only link to it the first time we reference it, unless we're specifically directing them to it somewhere else.

DannyFeliz:
> Good call. Would you mind just removing the second link as we typically only link to it the first time we reference it, unless we're specifically directing them to it somewhere else.

Done

jessarcher (approved):

PR #55, opened 2022-07-02
[0.3.x] Add `app/View/Components/**` to default refresh paths

timacdonald:
This PR adds `app/View/Components/**` to the global default directories to watch and perform full page reloads for. This is useful as View Components may be entirely specified in the class itself with no blade file.

Additionally, it also exports the default directories so users can utilise the default list and just add their own. Here would be an example of he export usage...

```js
import laravel, { refreshPaths } from 'laravel-vite-plugin';

export default defineConfig({
    plugins: [
        laravel({
            input: [
                'resources/css/app.css',
                'resources/js/app.js',
            ],
            refresh: [
                ...refreshPaths,
                'app/Http/Livewire/**'
            ],
        }),
    ],
});

```

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -105,6 +105,9 @@ export default function laravel(config: string|string[]|PluginConfig): [LaravelP
     ];
 }
 
+/**
+ * Resolve the Laravel Plugin configuration.
+ */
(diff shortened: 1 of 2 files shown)


jessarcher (approved):
Dig the export :)

jessarcher (approved):

PR #42, opened 2022-06-29
Use HMR host when set

jessarcher:
This PR updates the hot file generation to use Vite's [`server.hmr.host`](https://vitejs.dev/config/#server-hmr) config option when specified by the user.

This configuration option is intended to tell the Vite client where to find the HMR server for cases where it is different from [`server.host`](https://vitejs.dev/config/#server-host). I believe it makes sense for us to use it as well.

This probably helps a range of use cases, but specifically, it allows users to run Vite inside a Sail container that is running inside WSL, where the Vite dev server needs to be configured on the container's public address, but where the host machine cannot connect to this same address.

Closes #28

PR #56, opened 2022-07-02
Account for an hmr protocol different than the vite server's protocol

jbraband:
Currently, the protocol used for the URL presented to the client (for connecting to Vite dev server) is based on the `server.https` config of the dev server.

`const protocol = server.config.server.https ? 'https' : 'http'`

It is reasonable that a user would want the Vite dev server to run without SSL, but have the client connect using SSL. I use a reverse proxy container in my docker environment which routes traffic to the Laravel Sail containers for a couple different projects. The reverse proxy is responsible for terminating the SSL connection and it in turn communicates with the service container over HTTP.

In this environment, the Vite dev server is running and listening for HTTP from within the Laravel Sail container, but the client needs to communicate with the reverse proxy over HTTPS. 

`server.hmr.protocol` is pre-existing but is not considered by `laravel/vite-plugin` when building the URL for the client connection.

My approach piggy-backs on @jessarcher's PR #42 where the `server.hmr.host` config is allowed to override `server.host`

This PR is backwards compatible as omitting `server.hmr.protocol` will render a client URL using `server.https` to determine the protocol, as was being done before.

I ran the test suite before adding a test for this, but there are no current tests for the `viteDevServerUrl` variable and I could not figure out how to make assertions against its value.

I'm more than happy to provide a test if someone could help with a way to assert against the value of `viteDevServerUrl`

Please ask any questions about this, I can also provide additional environment details with the reverse proxy if that would help!

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -169,11 +169,16 @@ function resolveLaravelPlugin(pluginConfig: Required<PluginConfig>): LaravelPlug
 
                 const isAddressInfo = (x: string|AddressInfo|null|undefined): x is AddressInfo => typeof x === 'object'
                 if (isAddressInfo(address)) {
-                    const protocol = server.config.server.https ? 'https' : 'http'
+                    const configHmrProtocol = typeof server.config.server.hmr === 'object' ? server.config.server.hmr.protocol : null
+                    const clientHmrProtocol = configHmrProtocol ? (configHmrProtocol === 'wss' ? 'https' : 'http') : null
+                    const configProtocol = server.config.server.https ? 'https' : 'http'
+                    const protocol = clientHmrProtocol ?? configProtocol
+
                     const configHmrHost = typeof server.config.server.hmr === 'object' ? server.config.server.hmr.host : null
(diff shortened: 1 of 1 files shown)


jbraband:
maybe I should add that I am using this in my development environment for the past couple days and HMR is triggering properly through the socket connection 🎉

jessarcher:
Thanks, @jbraband! This makes sense to me. I'll give it a test soon and see if we can get it into the next release.

jessarcher:
Thanks again @jbraband - I've tested this locally and it works a treat. I just did a bit of refactoring as that section was starting to get a bit busy.

jbraband:
@jessarcher thanks for the refactor and the merge! 

I really dislike managing a fork with a needed change and for something like this, I'd just stay on Laravel Mix, which I have had my own issues with 🙄 

Vite is killer so far once I got the connection working with the reverse proxy.

Cheers! 🎉

jessarcher:
Awesome! We'll include this in today's scheduled release :+1:

PR #69, opened 2022-07-07
[0.x] Update issue template

timacdonald:
This PR asks users what operating system they are running, what web browser they are using, and if they are running in Sail or Docker, all of which are important when debugging issues.


jessarcher (approved):

timacdonald:
@driesvints just messaging you here to make sure you have eyes on it when you get back.

PR #73, opened 2022-07-09
Add note for people using Laravel Sail to get local dev environment working with Vite.

ammar-madni:
I recently followed the migration guide to migrate from Laravel mix to Vite. I completed all the steps and ran `sail npm run dev`. Unfortunately, it did not work for me, all I could see was a blank screen and some errors in the browser console. This was because I was using Laravel sail for local development. The issue was pretty simple and I thought it would be worth mentioning in the Upgrade guide.

Vite uses a different port (5173) for HMR compared to Laravel mix (8080), and since the default HMR port in Laravel sail is set to 8080, local development does not work without changing the port mapping in your `docker-compose.yml`.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -330,21 +330,19 @@ node storage/ssr/ssr.js
 
 ### Optional: Update default HMR port when using Laravel Sail
 
-If you are using Laravel Sail, you must update the `HMR_PORT` in your `.env` file.
-
-```diff
-- HMR_PORT=8080
-+ HMR_PORT=5173
-```
-
-Alternatively, you can set the default value in your `docker-compose.yml` file.
-
+If you are using Laravel Sail, you need to update the default `HMR_PORT` value and the port it maps to in your `docker-compose.yml` file.
 
 ```diff
 - -'${HMR_PORT:-8080}:8080'
 + -'${HMR_PORT:-5173}:5173'
 ```
 
+If you are setting the HMR_PORT in your `.env` file you will also need to update it to match.
+
+```diff
+- HMR_PORT=8080
++ HMR_PORT=5173
+```
 
 ### Wrapping up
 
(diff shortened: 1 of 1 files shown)


jessarcher:
Hi @ammar-madni, the `HMR_PORT` env var was only available in Sail for a short time as it had some issues and [was reverted](https://github.com/laravel/sail/pull/432).

The supported way to expose Vite in Sail is to use the `VITE_PORT` env var introduced in https://github.com/laravel/sail/pull/433. Note that this differs from the `HMR_PORT` because it puts the env var on both sides of the mapping. The Vite plugin will look for this env var and listen on the correct port, so that any links generated by Vite will also point to the correct port.

I'm happy to add instructions on adding the `VITE_PORT` binding, but I don't think it's worth mentioning `HMR_PORT` given it was so shortlived.

ammar-madni:
> Hi @ammar-madni, the `HMR_PORT` env var was only available in Sail for a short time as it had some issues and [was reverted](https://github.com/laravel/sail/pull/432).
> 
> The supported way to expose Vite in Sail is to use the `VITE_PORT` env var introduced in [laravel/sail#433](https://github.com/laravel/sail/pull/433). Note that this differs from the `HMR_PORT` because it puts the env var on both sides of the mapping. The Vite plugin will look for this env var and listen on the correct port, so that any links generated by Vite will also point to the correct port.
> 
> I'd be happy to merge this PR if it were updated to only document adding the new `VITE_PORT` binding.

Ok, that's great. So I should just remove the part about updating the `.env` file. And just demonstrate adding the `VITE_PORT` binding to the `docker-compose.yml` file instead of a diff?

ammar-madni:
@jessarcher I have made the changes requested. Please let me know if there is anything else I need to do.

jessarcher:
Thanks!

Issue #89, opened 2022-07-18
Consider not hardcoding `ASSET_URL` into the base path for production builds


localpath:
Introduce a way to override the base path used for production builds. In some instance the asset_url may be assigned to a different CDN or location than where the build assets need to be ie (a different CDN or the app server itself).

https://github.com/laravel/vite-plugin/blob/d0b8894b7fa8e4086d8c0d274b7cb26a7b93a5c1/src/index.ts#L100

https://github.com/laravel/vite-plugin/blob/d0b8894b7fa8e4086d8c0d274b7cb26a7b93a5c1/src/index.ts#L316

localpath:
Hello. Thanks for the quick response. So for instance the `ASSET_URL` is using a cdn for known static files but the build pipeline is not "building" the SPA/JS/LESS and then copying it over to a bucket to be picked up by the cdn for instance.

I like ur idea of the build step env var. That would work perfectly. My point was just that it forces the entire deployment to use or not use the native laravel asset helper as the base URL and I made a new temporary helper called `secure_cdn($path)` for instance.

I'll give the env at build step a shot next time unless we can merge on that variable in the laravel-vite plugin

driesvints:
It seems you sent in a PR for this, thanks. Let's continue the discussion there: https://github.com/laravel/vite-plugin/pull/91

PR #91, opened 2022-07-19
Use additional optional nullish env var for complicated asset url prop

localpath:
Use additional optional nullish env var for complicated asset url prop

See https://github.com/laravel/vite-plugin/issues/89

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -97,7 +97,7 @@ function resolveLaravelPlugin(pluginConfig: Required<PluginConfig>): LaravelPlug
         config: (userConfig, { command, mode }) => {
             const ssr = !! userConfig.build?.ssr
             const env = loadEnv(mode, userConfig.envDir || process.cwd(), '')
-            const assetUrl = env.ASSET_URL ?? ''
+            const assetUrl = env.VITE_ASSET_URL ?? env.ASSET_URL ?? ''
 
             return {
                 base: command === 'build' ? resolveBase(pluginConfig, assetUrl) : '',


localpath:
This was just my inclination as a quick way to provide an alternate asset url. Thinking about it, it could be very helpful bc say u have a CDN for all your apps and you upload this it would necessarily have to be the /assets/**/** dir in ur cdn. With an alternate you can build out to the local app server or an alternate CDN from what the Laravel asset helper is normally bound to.

timacdonald:
@localpath can you explain this one in more detail? It seems to me that `ASSET_URL` is the existing solution to what you are trying to achieve with this, so I don't understand how introducing another asset URL env adds any value.

I'm probably just missing something.

i.e. if you are hosting your assets on S3 in production you set `ASSET_URL` to point to your S3 bucket and everything should just work™️

When would I use this new proposed variable?

localpath:
Hi. So looking at the issue I realized that it really is the `Vite.php` in the framework that generatest the script tags that is the issue. It's using the `asset()` helper, so no way around that for me I guess. I think I'm stuck using a custom helper like `cdn_asset()` to point to the CDN and bind the app url in where `ASSET_URL` normally goes in `app.config`.

See here: https://github.com/laravel/framework/blob/4fb69abc9633ed1a3f67b4b37b24a9793fd67f83/src/Illuminate/Foundation/Vite.php#L57

njoguamos:
@localpath Today, I encountered the same problem.

I want to serve some images from CDN, but I want the js and CSS to be served from the build folder. 

Why? Because I don't want to push the build folder to CDN every time I deploy.

I ended up creating a custom `cdn_asset` helper.

shinnlu:
i have the same problem, but not for CDN.
i have two domain with 'admin.website.com' and 'website.com'
```
export default defineConfig({
	plugins: [
		laravel([
			'resources/assets/frontend/main.js',
			'resources/assets/backend/myApp/myApp.js'
		]),
		vue({
			template: {
				transformAssetUrls: {
					base: null,
					includeAbsolute: false,
				},
			},
		})
	]
});
```
the CORS error occur when i access admin.website.com, because the ASSET_URL is fixed.
```
Access to script at 'https://website.com/build/assets/myApp-95bdec68.js' from origin 'https://admin.website.com' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource.
```

PR #98, opened 2022-07-21
Prevent console warning when Vite pings HMR

### LINKED ISSUE #96: Error while using Inertia Vue: The Vite server should not be accessed directly.
- Laravel Vite Plugin Version: 0.5.0
- Laravel Version: 9.21.0
- Node Version: 16.15.0
- NPM Version: 8.14.0
- Host operating system: macOS 12.4
- Web Browser & Version: Chrome 103

### Description:
I have scaffolded an Inertia Vue SSR application using Laravel breeze with the following breeze command:

php artisan breeze:install vue --ssr

Everything works correctly but there is a little bug that bothers me, almost all the time, when i start the dev server i get the following warning in the console: The Vite server should not be accessed directly. Your Laravel application's configured APP_URL is: X

Im not accessing the vite dev server url, im using Laravel Valet and im accessing the "fancy" vhost url.

#### Install Laravel and enter the directory:
laravel new test && cd test

#### Require Breeze
composer require laravel/breeze

#### Scaffold a Breeze Inertia Vue SSR app using
php artisan breeze:install vue --ssr

#### Run the dev server
npm install && npm run dev

#### Go to your app
Using your web browser navigate to your app, usually the APP_URL that the dev server suggests

#### Stop the dev server
WITHOUT CLOSING YOUR APP TAB ON YOUR BROWSER Ctrl + c on the console, to close the dev server

#### Start it again
npm run dev

Now you should be seeing the warning, which shouldnt be showing because the user never accesed the dev server URL  directly

Just noticed this also happens scaffolding the Breeze Vue Inertia application without SSR.

jessarcher:
This PR removes the console warning when the dev server is accessed directly.

This addresses an issue with Vite 3 where it now pings the root URL when it loses connection with the HMR server, showing this misleading error message.

It's tricky to distinguish between a user visit and a ping, because they're both made by the browser. While we could do it, potentially with some upstream help, I don't know that it's worth the effort as the index page we serve should hopefully be enough to guide users towards accessing a Laravel dev server.

Fixes #96

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -184,12 +184,6 @@ function resolveLaravelPlugin(pluginConfig: Required<PluginConfig>): LaravelPlug
 
             return () => server.middlewares.use((req, res, next) => {
                 if (req.url === '/index.html') {
-                    server.config.logger.warn(
-                        "\n" + colors.bgYellow(
-                            colors.black(`The Vite server should not be accessed directly. Your Laravel application's configured APP_URL is: ${appUrl}`)
-                        )
-                    )
-
                     res.statusCode = 404
 
                     res.end(


timacdonald (approved):
I think this is fine. I did play with having the Vite client send a `VITE-PING` header so we could detect the difference, but I think we could look at it again and restore this in the future if it continues to be a problem even with the HTML warning page.

Issue #103, opened 2022-07-21
HRM hot file use IPv6 while vite devserver running on IPv4


jenky:
- Laravel Vite Plugin Version: 0.5.0
- Laravel Version: 9.21.4
- Node Version: 18.3.0
- NPM Version: 8.11.0
- Host operating system: macOS Monterey 12.4
- Web Browser & Version: Chome 103.0.5060.114 (arm64)
- Running in Sail / Docker: No (Valet)

### Description:
HRM hot file use IPv6 while vite devserver running on IPv4

<img width="1113" alt="image" src="https://user-images.githubusercontent.com/1808758/180252591-6d9f4a9e-871e-40e2-bfac-40cb184f9994.png">

The URL inside `public/hot` file: `http://::1:5173`

<img width="649" alt="image" src="https://user-images.githubusercontent.com/1808758/180252341-599ef298-f357-462e-a46d-2c52eb617d5c.png">

### Steps To Reproduce:
Upgrade from 0.4.0 and vite 2.9.9, then starts the dev server with `npm run dev`

jessarcher:
Hi @jenky, by default the Vite dev server listens on IPv4 and IPv6, if your OS supports it. 

I can't see anything in your issue that indicates that the Vite dev server is only listening on IPv4. `localhost` is a hostname and can resolve to both IPv6 (::1) and IPv4 (127.0.0.1) addresses.

jenky:
Hmm, but some how I can't visit http://::1:5173/@vite/client nor http://::1:5173/resources/js/app.js. Chrome decide to block it and redirect to blank page `about:blank#blocked`

jessarcher:
That's strange!

The plugin should be adding the brackets for IPv6 addresses:
https://github.com/laravel/vite-plugin/blob/6ed250de5cb325075ad539b415cb01488e28dd1c/src/index.ts#L342

jessarcher:
We've just double-checked this on macOS and the hot file has:

```
http://[::1]:5173
```

Do you have any `server` config in your `vite.config.js`?

jenky:
No I don't. This is my current `vite.config.js`

``` js
import { defineConfig, splitVendorChunkPlugin } from 'vite'
import vue from '@vitejs/plugin-vue'
import laravel from 'laravel-vite-plugin'
import DefineOptions from 'unplugin-vue-define-options/vite'
import AutoImport from 'unplugin-auto-import/vite'

export default defineConfig({
  plugins: [
[...]

PR #108, opened 2022-07-25
[0.5.x] Support both string and integer based IP versions

timacdonald:
Fixes: https://github.com/laravel/vite-plugin/issues/103

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -345,10 +345,19 @@ function resolveDevServerUrl(address: AddressInfo, config: ResolvedConfig): DevS
 
     const configHmrHost = typeof config.server.hmr === 'object' ? config.server.hmr.host : null
     const configHost = typeof config.server.host === 'string' ? config.server.host : null
-    const serverAddress = address.family === 'IPv6' ? `[${address.address}]` : address.address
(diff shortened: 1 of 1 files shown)


src/index.ts
```diff
@@ -345,12 +345,21 @@ function resolveDevServerUrl(address: AddressInfo, config: ResolvedConfig): DevS
...
     const configHmrHost = typeof config.server.hmr === 'object' ? config.server.hmr.host : null
     const configHost = typeof config.server.host === 'string' ? config.server.host : null
-    const serverAddress = address.family === 'IPv6' ? `[${address.address}]` : address.address
+    const serverAddress = isIpV6(address) ? `[${address.address}]` : address.address
     const host = configHmrHost ?? configHost ?? serverAddress
 
     return `${protocol}://${host}:${address.port}`
 }
 
+function isIpV6(address: AddressInfo): boolean {
```
jessarcher, line 354: I think `isIpv6` might read a bit better. I think generally initialisms in camel case are treated like they're a word. E.g. `isApi`

timacdonald: good point! Updated.

jessarcher (approved):

PR #113, opened 2022-08-03
Allow chosing different env files

LucdoOf:
Hello, 

I have a project in which i need to work with several env files, and none of them are simply called `.env`. 

I usually chose my env file using the `mode` option of vite.

But, when you run a vite server using `mode` option on an existing env file without a default `.env` file the server crash with this error: 

```
/var/www/Boreal/Upcover/Instances/Upcovel/node_modules/laravel-vite-plugin/dist/index.js:110
                        server.config.logger.info(`  ${picocolors_1.default.green('➜')}  ${picocolors_1.default.bold('APP_URL')}: ${picocolors_1.default.cyan(appUrl.replace(/:(\d+)/, (_, port) => `:${picocolors_1.default.bold(port)}`))}`);
                                                                                                                                                                     ^

TypeError: Cannot read properties of undefined (reading 'replace')
    at Timeout._onTimeout (/var/www/Boreal/Upcover/Instances/Upcovel/node_modules/laravel-vite-plugin/dist/index.js:110:166)
    at listOnTimeout (node:internal/timers:559:17)
    at processTimers (node:internal/timers:502:7)
```

In fact, in the `configureServer` method the env file loaded is the `.env` file, and it cannot be changed: 

```
const appUrl = loadEnv('', envDir, 'APP_URL').APP_URL
```

I suggest to change this line to:

```
const appUrl = loadEnv(resolvedConfig.mode, envDir, 'APP_URL').APP_URL
```

And thus, add:

```
mode: mode,
```

In the config definition, it fixes the issue.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -99,7 +99,6 @@ function resolveLaravelPlugin(pluginConfig: Required<PluginConfig>): LaravelPlug
             const assetUrl = env.ASSET_URL ?? ''
 
             return {
-                mode: mode,
                 base: command === 'build' ? resolveBase(pluginConfig, assetUrl) : '',
                 publicDir: false,
                 build: {
--- a/src/index.ts
+++ b/src/index.ts
@@ -99,6 +99,7 @@ function resolveLaravelPlugin(pluginConfig: Required<PluginConfig>): LaravelPlug
             const assetUrl = env.ASSET_URL ?? ''
 
             return {
+                mode: mode,
                 base: command === 'build' ? resolveBase(pluginConfig, assetUrl) : '',
                 publicDir: false,
                 build: {
@@ -147,7 +148,7 @@ function resolveLaravelPlugin(pluginConfig: Required<PluginConfig>): LaravelPlug
             const hotFile = path.join(pluginConfig.publicDirectory, 'hot')
 
             const envDir = resolvedConfig.envDir || process.cwd()
-            const appUrl = loadEnv('', envDir, 'APP_URL').APP_URL
+            const appUrl = loadEnv(resolvedConfig.mode, envDir, 'APP_URL').APP_URL
 
             server.httpServer?.once('listening', () => {
                 const address = server.httpServer?.address()
@@ -341,7 +342,7 @@ function resolveDevServerUrl(address: AddressInfo, config: ResolvedConfig): DevS
     const configHost = typeof config.server.host === 'string' ? config.server.host : null
     const serverAddress = isIpv6(address) ? `[${address.address}]` : address.address
     const host = configHmrHost ?? configHost ?? serverAddress
-    
+
     const configHmrClientPort = typeof config.server.hmr === 'object' ? config.server.hmr.clientPort : null
     const port = configHmrClientPort ?? address.port
 
--- a/src/index.ts
+++ b/src/index.ts
@@ -147,7 +147,7 @@ function resolveLaravelPlugin(pluginConfig: Required<PluginConfig>): LaravelPlug
             const hotFile = path.join(pluginConfig.publicDirectory, 'hot')
 
             const envDir = resolvedConfig.envDir || process.cwd()
-            const appUrl = loadEnv(resolvedConfig.mode, envDir, 'APP_URL').APP_URL
+            const appUrl = loadEnv(resolvedConfig.mode, envDir, 'APP_URL').APP_URL ?? 'undefined'
 
             server.httpServer?.once('listening', () => {
                 const address = server.httpServer?.address()


src/index.ts
```diff
@@ -99,6 +99,7 @@ function resolveLaravelPlugin(pluginConfig: Required<PluginConfig>): LaravelPlug
             const assetUrl = env.ASSET_URL ?? ''
 
             return {
+                mode: mode,
```
jessarcher, line 102: Was there a reason for this change?

jessarcher:
This makes sense. Not sure why I didn't do this originally, but I've given it a test using `npx vite -m example` and it works well :+1:

src/index.ts
LucdoOf: Oh really? Thought it wasn't defined at all. Maybe i went to quickly! Thanks for the quick review :)

jessarcher: If the user hasn't specified a mode (either via config or the `-m` flag) then it will be undefined in the `UserConfig` object at this point. Vite will automatically add the mode to the `ResolvedConfig` object that we access later, with a default of `development`, so there's no need for us to manually add it to the `UserConfig`, unless we needed to change it.

If you can remove this line, I'll get this merged in :)

LucdoOf: The thing is i don't want vite to parse ".env.development" but my own .env file. If i remove this line and no ".env.development" exists, the same error appears.
Even if i specify a mode using -m, without explicitly adding it to the ResolvedConfig it is set to 'development'.

jessarcher: Specifying `mode: mode` doesn't really do anything as far as I can tell. The `ResolvedConfig` object will have the same value with and without it being specified.

The default will always be `development` with or without it. If you specify a mode in your config or with `-m`, then the `ResolvedConfig` config should have that value, even without `mode: mode`.

If `.env.development` doesn't exist, that's okay. Vite looks for a few env files and merges the values together: https://github.com/vitejs/vite/blob/60721ac53a1bf326d1cac097f23642faede234ff/packages/vite/src/node/env.ts#L20-L25

Any files that don't exist are silently ignored.

The error you're seeing should only occur if none of the env files contains an `APP_URL`.

src/index.ts
```diff
@@ -147,7 +148,7 @@ function resolveLaravelPlugin(pluginConfig: Required<PluginConfig>): LaravelPlug
             const hotFile = path.join(pluginConfig.publicDirectory, 'hot')
 
             const envDir = resolvedConfig.envDir || process.cwd()
-            const appUrl = loadEnv('', envDir, 'APP_URL').APP_URL
+            const appUrl = loadEnv(resolvedConfig.mode, envDir, 'APP_URL').APP_URL
```
jessarcher, line 151: Something like this would prevent the error you're seeing:

```suggestion
            const appUrl = loadEnv(resolvedConfig.mode, envDir, 'APP_URL').APP_URL ?? 'undefined'
```

It would ensure that `appUrl` is a string when it doesn't exist in any of env files, which would prevent the `.replace()` call from failing.

The output would then look like this, which should hopefully make it obvious that the `APP_URL` couldn't be found.

image

We could throw an error if no `APP_URL` exists, but it's not required for this plugin to function.

src/index.ts
LucdoOf: Ok i understood the confusion.

I was using an npm script (npm run dev --mode=my_env_file) with the corresponding package.json:

```
"scripts": {
    "dev": "vite",
}
```

I don't know why, but when a mode is specified this way it doesn't appear in the resolvedConfig object. 

Running vite with vite --mode=my_env_file work tho. 

I'll just remove the mode:mode line, sorry for this !

src/index.ts
LucdoOf: I agree, would be a great idea. Just commited :)

src/index.ts
jessarcher: I don't think `npm run <script>` passes arguments or options through to the underlying command, so your `-m` flag would be ignored.

If you run `npm run dev -- -m mode` then it should pass everything after the `--` through.

jessarcher:
Looks good! Thanks for the contribution!

src/index.ts
LucdoOf: Thing is it get passed here correctly, even with the npm version:       

` config: (userConfig, { command, mode }) => {`

That's why i'm finding it kinda strange but its ok i'll deal with it :)

jessarcher: That is strange! When I run `npm run dev -m test`, the mode is passed into this function as `development`.

LucdoOf: Oh. Strange lol. Let's forget about this haha

PR #118, opened 2022-08-11
[0.5.x] Support customising hot file / build path

timacdonald:
This PR introduces a new configuration option to specify the location of the hot file. This is really just an escape hatch from our normal conventions.

```js
// vite.config.js

export default defineConfig({
    plugins: [
        laravel({
            hotFile: 'storage/vite.hot',
            // ...
        }),
    ],
});
```

```blade
// *.blade.php

{{ Vite::useHotFile(storage_path('vite.hot'))->withEntryPoints(['resources/js/app.js']) }}
```

This PR depends on: https://github.com/laravel/framework/pull/43620

The ability to customise the build directory (as seen in the above linked PR) is already possible without the need for additional changes to the plugin:

```js
// vite.config.js

export default defineConfig({
    plugins: [
        laravel({
            buildDirectory: 'foo/bar',
            // ...
        }),
    ],
});
```

```blade
// *.blade.php

{{ Vite::useBuildDirectory('foo/bar')->withEntryPoints(['resources/js/app.js']) }}
```

Note that by design the build directory is always relative to the public directory but the hot path is used as is.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -27,6 +27,8 @@ interface PluginConfig {
 
     /**
      * The path to the "hot" file.
+     *
+     * @default `${publicDirectory}/hot`
      */
     hotFile?: string
 
--- a/src/index.ts
+++ b/src/index.ts
@@ -26,6 +26,11 @@ interface PluginConfig {
     buildDirectory?: string
 
     /**
+     * The path to the "hot" file.
+     */
+    hotFile?: string
+
+    /**
      * The path of the SSR entry point.
      */
     ssr?: string|string[]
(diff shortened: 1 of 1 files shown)


src/index.ts
```diff
@@ -25,6 +25,11 @@ interface PluginConfig {
      */
     buildDirectory?: string
 
+    /**
+     * The path to the "hot" file.
+     */
+    hotFile?: string
```
jessarcher, line 31: Should we add a `@default` annotation to this one like the others?

It's a little tricky because it's partially computed, but I think `'public/hot'` would still make sense.

timacdonald: I did consider this at the time, but decided against it because it is calculated. However, I've compromised...and potentially broken the lawy, but specifying the default value as a computed value e.g. `${publicDirectory}/hot`

timacdonald:
Documentation PR: https://github.com/laravel/docs/pull/8146

PR #123, opened 2022-08-19
Update Vapor helper version

timacdonald:

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -246,7 +246,14 @@ composer require laravel/vapor-cli:^1.40.0
 
 Next, if you are using the Vapor asset helper in your application, you only need to utilize the asset helper when you are referencing assets you don't want bundled, such as those that already live in your public directory.
 
-If you want to use the asset helper with your Vite project, you will also need to specify the base URL for assets in your application's entry point, for example in your `resources/js/app.js,` like so:
+If you want to use the asset helper with your Vite project, you will also need to ensure you have updated to the latest version:
+
+
+```sh
+npm install laravel-vapor@^0.6.0
+```
+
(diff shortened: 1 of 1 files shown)


UPGRADE.md
```diff
@@ -246,7 +246,14 @@ composer require laravel/vapor-cli:^1.40.0
 
 Next, if you are using the Vapor asset helper in your application, you only need to utilize the asset helper when you are referencing assets you don't want bundled, such as those that already live in your public directory.
 
-If you want to use the asset helper with your Vite project, you will also need to specify the base URL for assets in your application's entry point, for example in your `resources/js/app.js,` like so:
+If you want to use the asset helper with your Vite project, you will also need to ensure you have updated to the latest version:
+
+
+```sh
+npm install laravel-vapor@^0.6.0
```
jessarcher, line 253: I think this would work and be less maintenance:

```suggestion
npm install laravel-vapor@latest
```

UPGRADE.md
```diff
@@ -246,7 +246,14 @@ composer require laravel/vapor-cli:^1.40.0
 
 Next, if you are using the Vapor asset helper in your application, you only need to utilize the asset helper when you are referencing assets you don't want bundled, such as those that already live in your public directory.
 
-If you want to use the asset helper with your Vite project, you will also need to specify the base URL for assets in your application's entry point, for example in your `resources/js/app.js,` like so:
+If you want to use the asset helper with your Vite project, you will also need to ensure you have updated to the latest version:
```
jessarcher, line 249: What do you think of this? Or are we expecting that they will already have it installed if they're reading this?

```suggestion
If you want to use the asset helper with your Vite project, you will also need to ensure you have installed the latest version:
```

timacdonald: I believe we approach the upgrade guide with the thought that they already had an existing project, so not adding anything new.

PR #129, opened 2022-08-25
[0.6.x] Add config option to utilise Valet TLS certificates

timacdonald:
This PR introduces a new `valetTls` configuration option to allow developers to use Valet generated certificates on their Mac. This is useful as Vite 3 no longer generates a self signed certificate. Valet also puts certificates into the system keychain, so if you are using Chromium or Safari the experience is seamless. 

Unfortunately Firefox does not seem to read from the system keychain, so you may still need to additionally accept the Valet certificate for the Vite server, as we currently have to.

When setting the configuration option to `true`, the plugin will automatically detect the host by checking the current directory name and appending the configured Valet TLD:

```js

export default defineConfig({
    plugins: [
        laravel({
            valetTls: true,
        }),
    ],
});
```

If you have linked you site via a different host, that is if you are in the `my-app` directory and ran:

```
valet link my-host
valet secure my-host
```

so that you can view "my-app" by visiting https://my-host.test, then you should specify the host along with the TLD in the `valetTls` option:

```js

export default defineConfig({
    plugins: [
        laravel({
            valetTls: 'my-host.test',
        }),
    ],
});
```

User config will always take precedence over the underlying generated Valet config.

Documentation PR: https://github.com/laravel/docs/pull/8164

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -429,7 +429,7 @@ function resolveValetServerConfig(host: string|boolean): {
     const certPath = path.resolve(os.homedir(), `.config/valet/Certificates/${host}.crt`)
 
     if (! fs.existsSync(keyPath) || ! fs.existsSync(certPath)) {
-        throw Error('Unable to find Valet certificate files. Ensure you have run `valet secure`.')
+        throw Error(`Unable to find Valet certificate files for your host [${host}]. Ensure you have run "valet secure".`)
     }
 
     return {
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,5 +1,6 @@
 import fs from 'fs'
 import { AddressInfo } from 'net'
+import os from 'os'
 import path from 'path'
 import colors from 'picocolors'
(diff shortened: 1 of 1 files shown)


jessarcher:
Looks good!

I note that Vite has introduced a separate plugin for SSL (https://github.com/vitejs/vite-plugin-basic-ssl) but I like this option for Valet users so they don't need to install anything additional.

I wonder whether we should document https://github.com/vitejs/vite-plugin-basic-ssl for non-Valet users?

timacdonald:
Interesting! I've added that to the docs, but kept it as the last line of defense as Vite is recommending people generate their own - but still good to include.

Screen Shot 2022-08-29 at 1 50 36 pm

jessarcher (approved):
Works a treat! :ship:

driesvints:
@jessarcher @timacdonald think @taylorotwell is not going to review this, right? Can you merge this yourself?

Issue #140, opened 2022-09-12
Laravel Vite page return 404 after restarting the server


amlxv:
- Laravel Vite Plugin Version: 0.6.0
- Laravel Version: 9.29.0
- Node Version: 18.7.0
- PNPM Version: 7.10.0
- Host operating system: Windows 11 Home Single Language 10.0.22000
- Web Browser & Version: Chrome 105.0.5195.102
- Running in Sail / Docker: No

### Description:
I have tried reproducing using Vite+TS, but the page still works even if the server is automatically restarted. 

![Screenshot 2022-09-12 230948](https://user-images.githubusercontent.com/34534953/189697003-0fe93746-ee1d-41e4-957e-741f7671cd08.png)

![Screenshot 2022-09-12 231050](https://user-images.githubusercontent.com/34534953/189697107-104263fc-ebad-4d6d-975f-096bc5f30a2a.png)

### Steps To Reproduce:
1. Create a Laravel project via the Composer.
2. Run `pnpm install` & `pnpm run dev`.
3. Open the given link (e.g., http://localhost:5173/).
4. The "Laravel Vite" page will normally appear.
5. Save the `vite.config.js`. It'll restart the server.
6. Revisit the link or refresh the "Laravel Vite" page will result in HTTP ERROR 404
7. Re-running  `pnpm run dev` can fix this problem but it still does not work when the server is automatically restarted.

timacdonald:
Thanks for letting us know.

Fixed with https://github.com/laravel/vite-plugin/pull/141

PR #141, opened 2022-09-13
[0.6.x] Ensure custom 404 page shows after server restarts

timacdonald:
fixes: #140

Currently the custom 404 page stops displaying if the server has restarted. This fixes that issue.


jessarcher (approved):
Didn't realise this was an issue - looks like the right fix!

PR #151, opened 2022-09-28
[0.6.x] Support loading certificates from environment variables

innocenzi:
This PR adds support for loading certificates thanks to paths defined in the environment. 

Currently, in our team, we use `valetTls`, which is very nice for us Valet users. But a member of our team uses Windows and Laragon, and cannot benefit from the configuration. 

With this PR, they will be able to define `VITE_DEV_SERVER_KEY` and `VITE_DEV_SERVER_CERT` in their `.env` with the paths to their Laragon certificates. That way, there is no need to touch `vite.config.ts`, so it can keep being environment-agnostic.

~~Keeping this as a draft to test it on my own Windows machine later~~ - also, not so sure about the variable names: `VITE_*` environment variables are available in the front-end through `import.meta.env`.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -117,6 +117,11 @@ function resolveLaravelPlugin(pluginConfig: Required<PluginConfig>): LaravelPlug
             const valetServerConfig = command === 'serve'
                 ? resolveValetServerConfig(pluginConfig.valetTls)
                 : undefined
+            const https = userConfig.server?.https === false ? false : {
+                ...valetServerConfig?.https,
+                ...resolveHttpsConfig(env),
+                ...(userConfig.server?.https === true ? {} : userConfig.server?.https),
+            }
 
             ensureCommandShouldRunInEnvironment(command, env)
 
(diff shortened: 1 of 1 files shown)


innocenzi:
This pull request is ready for review. If you have a better name for `VITE_DEV_SERVER_KEY`/`CERT`, feel free to suggest.

timacdonald:
Seems good to me! I've added some error messages to hopefully guide the user when we cannot find the specified paths, instead of silently failing...


Screen Shot 2022-10-11 at 10 33 28 am

timacdonald (approved):

timacdonald:
Thanks for contributing to the core plugin @innocenzi ❤️ you rock.

timacdonald:
@innocenzi I'm thinking we should be getting the value from the `ASSET_URL` first and then falling back to the `APP_URL`.

The asset URL is the definitive url that would be used by the asset helper as well, where as the app url may not be correct and isn't used by the asset helper.

What do you think?

timacdonald:
@innocenzi I'm wondering if we should first use the `ASSET_URL` and then fallback to the `APP_URL` here?

What do you think?

The official plugin uses the `asset` helper for all generated URLs.

innocenzi:
Admittedly I never used `ASSET_URL` in development myself so I'm not too sure. I imagine it could make sense when using Docker but I don't know the setup :/ 

I would say, we should probably wait for people with such setups to provide feedback if what the plugin offers doesn't work for them

PR #153, opened 2022-10-07
Vue 3

Stefanosapk:
Mention in the migration guide that besides Laravel 9.19, Vue 3 it is needed as well. Also in Laravel docs is not mentioned somewhere. Also if there is support for Vue 2, please make it crystal clear the differences in migration.


timacdonald:
The version of Vue, React,  etc you are using is dependent on your application. The plugin should work fine with either version and the given plugins are there for example.

PR #163, opened 2022-11-14
[0.8.x] Change certificate resolution

### LINKED ISSUE #162: `valetTls` should not be an option, certificates should be inferred from the environment
## Discussion

I'm opening this issue to find a truly environment-agnostic solution to loading certificates. For reference, my implementation works in all three of the situations described in the context below.

https://github.com/innocenzi/laravel-vite/blob/main/vite-plugin-laravel/src/config.ts#L245

I think this should be implemented this way in this plugin, but that would be a breaking change ~~and I don't want to PR something that will not get merged~~ (I PR'd anyway lol).

The solution in my eyes would be to remove `valetTls`, and try to find the certificates from the environment: 
- If `VITE_DEV_SERVER_KEY`/`VITE_DEV_SERVER_CERT` is provided, use that
- If not, try to load Valet certificates (from `APP_URL` and not from the current working directory)
- If none of the above, do not throw an error

## Context

We have a team where two developers are using macOS and one is using Windows. We used my own Vite integration, but going forward I'd like to use the official one. 

My integration is environment-agnostic and has the ability to detect certificates from both Valet and Laragon. To work around this missing feature, I PR'd https://github.com/laravel/vite-plugin/pull/151, which introduces support for loading certificates through environment variables. 

But today, when the other macOS developer tried to start the server, he got an exception (Valet certificates not found), and we realized his project directory was not the same as his app URL (which makes sense in our context, the project powers two different apps).

When trying to solve this issue, we simply realized `valetTls` should not be an option because it's not environment-agnostic. Worse than that, my previous PR only works if `valetTls` is not set to `true`, because otherwise an exception will be thrown when its certificates will not be found, regardless of the existence of the certificates provided by the `.env`.

innocenzi:
Closes #162

This pull request changes how the certificates for the development server are resolved. 

Previously, the `valetTls` option needed to be defined for Valet certificates to be inferred. However, this is not an environment-agnostic solution: if a developer works on Windows, they will get an error when the server starts.

I tried to mitigate this through https://github.com/laravel/vite-plugin/pull/151, but this requires `valetTls` to be removed/set to `false`, and everyone to specify their certificates in their `.env`, which is super inconvenient.

Even when using `valetTls`, the current implementation tries to find certificates by using the current working directory's name instead of the `APP_URL`, which is problematic when the directory name is not the same as the `APP_URL`. This can be fixed by setting `valetTls` to the name of the domain, but then this breaks the configuration for the rest of the team — so we're back to setting `valetTls` to `false` and manually specifying the certificates in the `.env`.

To fix all of these issues:
- The `valetTls` option in `vite.config.ts` should be removed
- Certificates should be inferred in a logical order (environment-defined certificates first, then Valet certificates)
- Certificate resolution should not throw an error

~~Marking this as draft to properly test it in all of our environments.~~

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -55,7 +55,7 @@ interface PluginConfig {
 
     /**
      * Utilise the valet TLS certificates.
-     *
+     * @deprecated
      * @default false
      */
     valetTls?: string|boolean,
@@ -115,7 +115,7 @@ function resolveLaravelPlugin(pluginConfig: Required<PluginConfig>): LaravelPlug
             const env = loadEnv(mode, userConfig.envDir || process.cwd(), '')
             const assetUrl = env.ASSET_URL ?? ''
             const serverConfig = command === 'serve'
-                ? (resolveValetServerConfig(pluginConfig.valetTls) ?? resolveEnvironmentServerConfig(env))
+                ? (resolveEnvironmentServerConfig(env) ?? resolveValetServerConfig(env))
                 : undefined
 
             ensureCommandShouldRunInEnvironment(command, env)
@@ -447,12 +447,8 @@ function resolveEnvironmentServerConfig(env: Record<string, string>): {
     host?: string,
     https?: { cert: Buffer, key: Buffer }
 }|undefined {
-    if (! env.VITE_DEV_SERVER_KEY && ! env.VITE_DEV_SERVER_CERT) {
-        return
-    }
-
     if (! fs.existsSync(env.VITE_DEV_SERVER_KEY) || ! fs.existsSync(env.VITE_DEV_SERVER_CERT)) {
-        throw Error(`Unable to find the certificate files specified in your environment. Ensure you have correctly configured VITE_DEV_SERVER_KEY: [${env.VITE_DEV_SERVER_KEY}] and VITE_DEV_SERVER_CERT: [${env.VITE_DEV_SERVER_CERT}].`)
+        return
     }
 
     const host = resolveHostFromEnv(env)
@@ -471,38 +467,21 @@ function resolveEnvironmentServerConfig(env: Record<string, string>): {
     }
 }
 
-/**
- * Resolve the host name from the environment.
- */
-function resolveHostFromEnv(env: Record<string, string>): string|undefined
-{
-    try {
-        return new URL(env.APP_URL).host
-    } catch {
-        return
-    }
-}
-
 
 /**
  * Resolve the valet server config for the given host.
  */
-function resolveValetServerConfig(host: string|boolean): {
+function resolveValetServerConfig(env: Record<string, string>): {
     hmr?: { host: string }
     host?: string,
     https?: { cert: Buffer, key: Buffer }
 }|undefined {
-    if (host === false) {
-        return
-    }
-
-    host = host === true ? resolveValetHost() : host
-
+		const host = resolveHostFromEnv(env)
     const keyPath = path.resolve(os.homedir(), `.config/valet/Certificates/${host}.key`)
     const certPath = path.resolve(os.homedir(), `.config/valet/Certificates/${host}.crt`)
 
-    if (! fs.existsSync(keyPath) || ! fs.existsSync(certPath)) {
-        throw Error(`Unable to find Valet certificate files for your host [${host}]. Ensure you have run "valet secure".`)
+    if (! host || ! fs.existsSync(keyPath) || ! fs.existsSync(certPath)) {
+        return
     }
 
     return {
@@ -516,16 +495,13 @@ function resolveValetServerConfig(host: string|boolean): {
 }
 
 /**
- * Resolve the valet valet host for the current directory.
+ * Resolve the host name from the environment.
  */
-function resolveValetHost(): string {
-    const configPath = os.homedir() + `/.config/valet/config.json`
-
-    if (! fs.existsSync(configPath)) {
-        throw Error('Unable to find the Valet configuration file. You will need to manually specify the host in the `valetTls` configuration option.')
+function resolveHostFromEnv(env: Record<string, string>): string|undefined
+{
+    try {
+        return new URL(env.APP_URL).host
+    } catch {
+        return
     }
-
-    const config: { tld: string } = JSON.parse(fs.readFileSync(configPath, 'utf-8'))
-
-    return path.basename(process.cwd()) + '.' + config.tld
 }
--- a/src/index.ts
+++ b/src/index.ts
@@ -476,7 +476,7 @@ function resolveValetServerConfig(env: Record<string, string>): {
     host?: string,
     https?: { cert: Buffer, key: Buffer }
 }|undefined {
-		const host = resolveHostFromEnv(env)
+    const host = resolveHostFromEnv(env)
     const keyPath = path.resolve(os.homedir(), `.config/valet/Certificates/${host}.key`)
     const certPath = path.resolve(os.homedir(), `.config/valet/Certificates/${host}.crt`)
 


timacdonald:
Hey Enzo!

I don't really see any issues with the current implementation.

- `valetTls` works as expected. It is a good feature that I personally use every day. I don't think we should remove it.
- I don't think we should ever use the `APP_URL` to determine the host. I think we need to reconsider how we did that in the env setup PR anyhow, I just haven't gotten to it.
- The valet setup determines your host in exactly the same way that valet itself determines the host. Alternatively, if you use a custom host with valet, you may configure it with  `valetTls: 'foo.test'`.
- `valetTls` should take precedence over the ENV config. It is more explicit than specifying things in the ENV. I think the current waterfall of config makes sense and personally wouldn't want it to work in reverse.
- I don't see how silently failing when the certificates you have specified helps? If you have specified a path and we can't find it, the developer facing errors are rather useful IMO.

Would it not be possible to just remove `valetTls` from your own projects config and use the ENV method instead?

Please let me know if I'm missing something though.

innocenzi:
Hey Tim!

Thanks for your work on this feature and for taking the time to review this. For the record, I added way more context in [this issue](https://github.com/laravel/vite-plugin/issues/162).

There is nothing wrong with the feature itself obviously, the issue is that its implementation is not environment agnostic.

> `valetTls` works as expected as far as I can tell. It is a good feature that I personally use every day. I don't think we should remove it unless there is something broken in the feature itself.

Yes, this is a good feature and it works. My suggestion is to remove the setting, not the feature, because of its non-agnostic nature.

> The valet setup determines your host in exactly the same way that valet itself determines the host.

I think having the host determined by the `.env` is better because that's the one the developer would use. We're 100% certain that's the one being used, as opposed to the current directory which could have a different name. This is the issue we're experiencing.

> Alternatively, if you use a custom host with valet, you may configure it with valetTls: 'foo.test'.

While this is true, modifying `vite.config.ts` this way would break the setup for every other developer in the team. We don't necessarily use the same domain name.

> `valetTls` should take precedence over the ENV config. It is more explicit than specifying things in the ENV. I think the current waterfall of config makes sense and personally wouldn't want it to work in reverse.

Anything in the configuration should take precedence over the environment, I agree. But this specific feature is not environment agnostic, that's why I'm suggesting an alternative that makes sense in a multi-member, multi-environment team, and still works for individuals in a transparent way.

> Would it not be possible to just remove valetTls from your own projects' config and use the ENV across the whole team instead?

While this is possible, I think it's unfortunate to give up the DX of certificate discovery. I don't know by heart the paths to my certificates and I don't want to have to set it up for each project that we work on as a team, for each member.

> I don't see how silently failing when the certificates can't be found helps? If you have specified a path and we can't find it, the developer facing errors are rather useful IMO.

To be fair, this is just the only drawback of my suggestion — since the certificate discovery is automatic, we can't know for sure the lack of certificate is an issue. For instance, if the developer doesn't use `https` locally, they still want the server to start without an exception.

Let me know if I missed some of your points or if I didn't explain well enough.

HassanZahirnia:
I've tested this PR in a Windows machine running WSL2 with no Valet and it works just fine! ✅👍

This is something we've struggled with since some members use Valet and I personally don't.
With this PR we can all work without having to worry about the `valetTls` being set 🚀
I hope this gets merged so this config becomes environment agnostic 🙌

driesvints:
FYI: draft pr's don't get reviewed.

innocenzi:
@driesvints Thanks—I've just marked this PR as ready for review, the person as was waiting for in my team to test this PR is still not available, but since it's been tested on Windows/without Valet, I think it's ready.

timacdonald:
After chatting over this one with @jessarcher yesterday, we think we want to hold off merging.

We want to keep the Laravel plugin focused on making Laravel work with Vite. Telling Vite about where generic TLS certificates exist on your system isn't really a "Laravel" problem and is already something that Vite supports out of the box.


Additionally, it seems to us that we have exposed all the tools needed for an array of team configurations:

- A solo dev using Laravel Valet  can easily just add `valetTls` to their config (we of course want first party support for Valet).
- A team exclusively using Valet may do the same.
- A team not using Valet at all may specify their certificate paths via Vite's first party support. They may hard-code paths or use ENV values.
- A team with a mixture of Valet and non-Valet devs may either:
  - Specify their certificate paths via Vite's first party support using ENV values.
  - Use the `valetTls` option for those that want it or use ENV values for those that do not.

The following is an example of how those using Valet may take advantage of auto-detection, while others not using Valet may specify the paths in their ENV.

```js
import fs from 'fs'
import { defineConfig, loadEnv } from 'vite';

const env = loadEnv('', '', '')
const host = new URL(env.APP_URL).host

export default defineConfig({
    plugins: [
        laravel({
            refresh: true,
            valetTls: ! env.VITE_DEV_SERVER_KEY
        }),
    ],
    ...env.VITE_DEV_SERVER_KEY ? {
        server: {
            hmr: { host },
            host,
            https: {
                key: fs.readFileSync(env.VITE_DEV_SERVER_KEY),
                cert: fs.readFileSync(env.VITE_DEV_SERVER_CERT),
            },
        },
    } : undefined,
});
```

innocenzi:
I don't understand the rationale :(

This change doesn't make the code more complex (it even removes code), and it keeps existing logic and even improves it by handling more scenarios. I think this is what makes a framework, especially Laravel: flexibility and developer experience.

> Additionally, it seems to us that we have exposed all the tools needed for an array of team configurations

While this is true, the whole point is that it is inconvenient. For mixed-environment teams, using the first solution, you don't get the benefit of automatic Valet certificate detection, and using the second one, you have to update your config with additional logic and conditions. 

All of this can easily be improved with this pull request with no specific maintenance cost.

I want to point out that the plugin isn't even necessary to make Vite works with Laravel in the first place, people have been doing it since Vite 2. It's really just about the developer experience, the reduced boilerplate.

I kindly ask you to reconsider your decision.

timacdonald:
@innocenzi circling back to this to get this wrapped up. My understanding is the following satisfies your requirements without needing an complex config

```js
import { defineConfig, loadEnv} from 'vite'
import laravel from 'laravel-vite-plugin'

const env = loadEnv('', '')

export default defineConfig({
  plugins: [
    laravel({
      input: 'resources/js/app.js',
      detectTls: env.VITE_DETECT_TLS === 'true' ? true : env.VITE_DETECT_TLS,
    }),
  ],
})
```

Then in your `.env`...

```
# file .env

# MacOS w/ Herd or Valet using conventional directory structure...
VITE_DETECT_TLS=true

# MacOS w/ Herd or Valet using unconventional directory structure...
VITE_DETECT_TLS=my-app.test

# Users not using Herd or Valet...
APP_URL=my-app.test
VITE_DEV_SERVER_KEY=/path/to/key
VITE_DEV_SERVER_CERT=/path/to/cert
```

## How certificate resolution currently works

Just documenting this for myself more than anything.

## detectTls

- When `detectTls` is `true` than we automatically resolve the certificate via Herd or Valet using `$CWD.{tld}`
- When `detectTls` is a string than we automatically resolve the certificate via Herd or Valet using `$detectTls.{tld}`

In both of these situations, an error is thrown when the certificate is not found.

- When `detectTls` is `false|undefined` we would attempt to resolve the certificate via the environment using the following...

## Environment resolution

- When `VITE_DEV_SERVER_KEY` or `VITE_DEV_SERVER_CERT` is not set, no TLS configuration is done.
- When either of the paths specified in `VITE_DEV_SERVER_*` do not exist an error is thrown.
- When a host cannot be found via the `APP_URL` environment variable an error is thrown.
- Otherwise the host and certificates are used to configure HTTPS.

timacdonald:
I think the only change I would want to make here it to have the Valet / Herd certificates auto detection in place, which we are looking at in https://github.com/laravel/vite-plugin/pull/180

innocenzi:
Thank you for taking yet another look! I appreciate you spending time on this.

Just like the [previous suggestion](https://github.com/laravel/vite-plugin/pull/163#issuecomment-1332871983), while yes, this would work, it is not ideal and does not improve the developer experience as I was hoping.

As stated in my other PR, I'll instead use my own plugin to fill my apparently specific needs, so I'm going to close this one. Thank you 👍

PR #166, opened 2022-11-17
[0.7.x] Fix node type issue

timacdonald:
Fixes this issue we have with the build by upgrading the version of typescript we are using.

Upgraded `vitetest` while I was in there.

Screen Shot 2022-11-18 at 10 01 19 am


driesvints:
Cool, Thanks!

PR #169, opened 2022-11-19
Allow hot file host to be configured

nurdism:
In my dev container solution I am running the whole thing behind a reverse proxy and proxying `/@vite*`, `/node_modules*` & `/resources*` to the running vite instance. I've been running into is not being able to force this plugin to either not specify a host or to force it to use a host. This option is optional and allows you to specify a host rather than attempting to figure it out.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -59,6 +59,13 @@ interface PluginConfig {
      * @default false
      */
     valetTls?: string|boolean,
+    
+    /**
+     * Override the host for the "hot" file.
+     * 
+     * @default false
+     */
+    host?: string|boolean,
 }
 
 interface RefreshConfig {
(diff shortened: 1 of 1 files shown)


timacdonald:
Hi @nurdism,

Thanks for the PR.

You can completely control the hot file by specifying the server configuration. The following shows you how to control the protocol, host, and port that ends up in the hot file.

```js

export default defineConfig({
    plugins: [
        laravel({
            refresh: true,
        }),
    ],
    server: {
        https: true,
        host: 'foo.test',
        port: '8080',
    }
});
```

I understand that you may have got this sorted and are looking to make it easier, however I personally would prefer to not add another way to control this - at least for now.

Thanks again.

nurdism:
@timacdonald adding 

```
server: {
        https: true,
        host: 'foo.test',
        port: '8080',
}```

does not solve this issue. I'm trying to force no host on the hot file. I want the host to be blank so it loads from the reverse proxy your [function](https://github.com/laravel/vite-plugin/blob/main/src/index.ts#L405) returns a hard coded `${protocol}://${host}:${port}`. Its running behind a reverse proxy on a different port, it also does not resolve the host that way, it resolves to `https://[::]:8080`

PR #170, opened 2022-11-21
Update RELEASE.md

driesvints:
Makes sure the release process is documented. Should we also do this for Echo? 

Ideally we'd automate this.


jessarcher (approved):
Looks good to me - the only thing that's potentially missing is the need to have run `npm login` on the machine at least once to authenticate with npmjs.com, but maybe that's a given.

driesvints:
Yeah think that's a given.

timacdonald (approved):

PR #173, opened 2022-11-25
Serve assets from the public directory during dev

### LINKED ISSUE #171: Vite publicDir does not copy and serve files correctly
- Laravel Vite Plugin Version: 0.7.1
- Laravel Version: 9.41.0
- Node Version: 18.8.0
- NPM Version: 8.18.0
- Host operating system: Windows / Linux / macOS Windows 10
- Web Browser & Version: Firefox/Chrome?

### Description:
Vite does not copy files from publicDir folder when running npm run build.

### Steps To Reproduce:
- In an standar and clean installation of Laravel 9.41.0
- Create a in /resources a folder called /assets
- Enable publicDir in vite.config.js
```

export default defineConfig({
    publicDir: 'resources/assets',
    plugins: [
        laravel({
            refresh: true,
        }),
    ],
});
```
- run `npm run build`
- According to the Vite documentation (https://vitejs.dev/config/shared-options.html#publicdir), all files are served in `/` during development and copied to the root of `outDir` during bulding ( also known as a public folder). But none of this has happened. Files are not copied or server-side when referenced from any page or css url(). The output is:

image

- Now I make the copies of the public folder explicit, and I specify `outDir` in `vite.config.js`
```

export default defineConfig({
    publicDir: 'resources/assets',
    plugins: [
        laravel({
            refresh: true,
        }),
    ],
    build: {
        sourcemap: false,
        emptyOutDir: true,
        copyPublicDir: true,    
        outDir: 'dist'
    }
});
```
- run `npm run build` and the ouput is:
 
image

I need put in this folder some assets like images, pdf, fonts, etc. and I don't want Vite to process and versioning them.

jessarcher:
This PR addresses an issue during dev mode when referencing absolute paths in CSS when not importing the CSS via JS (e.g. when using Blade).

In this scenario, the CSS file is loaded directly from the Vite dev server (e.g. `http://127.0.0.1:5173/resources/js/app.css`) rather than it being injected into the head of the page being served by Laravel (e.g. `http://127.0.0.1:8000`).

Any absolute paths in the CSS are fetched from the host and port serving the CSS (e.g. `http://127.0.0.1:5173/images/taylor.jpg`) which the Vite dev server currently can't resolve.

This PR addresses this by setting Vite's [`publicDir`](https://vitejs.dev/config/shared-options.html#publicdir) config to the public directory *during dev mode only* so that it may resolve and serve the assets.

Importantly, the setting is left `false` during build because Vite's default behaviour is to copy everything from this directory into the `outDir`, which in our case is a child of the public directory.

This is not an issue when in build mode because the CSS is served by the same web server as the rest of the application.

Fixes #171

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -122,7 +122,7 @@ function resolveLaravelPlugin(pluginConfig: Required<PluginConfig>): LaravelPlug
 
             return {
                 base: command === 'build' ? resolveBase(pluginConfig, assetUrl) : '',
-                publicDir: false,
+                publicDir: command === 'build' ? false : pluginConfig.publicDirectory,
                 build: {
                     manifest: userConfig.build?.manifest ?? !ssr,
                     outDir: userConfig.build?.outDir ?? resolveOutDir(pluginConfig, ssr),
--- a/tests/index.test.ts
+++ b/tests/index.test.ts
@@ -213,6 +213,20 @@ describe('laravel-vite-plugin', () => {
         delete process.env.LARAVEL_SAIL
     })
 
+    it('serves assets from the public directory during dev', () => {
+        const plugin = laravel('resources/js/app.js')[0]
+
+        const config = plugin.config({}, { command: 'serve', mode: 'development' })
+        expect(config.publicDir).toBe('public')
+    })
+
+    it('does not copy the public directory to the outDir on build', () => {
+        const plugin = laravel('resources/js/app.js')[0]
+
+        const config = plugin.config({}, { command: 'build', mode: 'production' })
+        expect(config.publicDir).toBe(false)
+    })
+
     it('prevents the Inertia helpers from being externalized', () => {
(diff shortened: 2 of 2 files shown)


jessarcher:
Marking this as draft until I've tested it further.

jessarcher:
I'm also thinking this plugin should respect _everything_ from `userConfig`. I don't like that the changes in this PR are impossible to make in user land.

timacdonald:
- Does it use it as a fallback or priority. `./resources/js/app.js` and `./public/resources/js/app.js`

jessarcher:
Going to close this one for the time being as the behaviour is perhaps a little too magical and may lead developers to think something will work in production that may not.

Instead, #176 will allow this behaviour to be enabled manually via `vite.config.js` like so:

```js
export default defineConfig(({ command }) => ({
  publicDir: command === 'serve' ? 'public' : false,

  // ...
}))
```

Based on feedback we can look at this again in the future.

PR #180, opened 2022-12-09
[1.x] Auto detect Valet / Herd TLS certificates

timacdonald:
This PR adds:

- Valet / Herd certificate auto-detection (✨) and makes it the default for upcoming `1.x` release.
- Ability to opt-out of auto-detection `detectTls: false`.
- CLI output to indicate when the Valet / Herd certificate is being used (when known).

Assume everywhere "Valet" is mentioned that it also applies to "Herd".

This PR allows teams with a mixture of Valet and non-Valet users to configure TLS certificates, without adding additional configuration.

The following standard Laravel plugin config supports both use-cases:

```js
import fs from 'fs'

export default defineConfig({
    plugins: [
        laravel({
            refresh: true,
        }),
    ],
});
```

For Valet users, as long as a Valet certificate is found it will be used. Valet certificates are "guessed" in the same way they currently are.

1. $CWD is used as the host.
2. TLD is resolved from the Valet config.

For non-Valet users, you may specify the path to TLS certificates with the existing environment options.

```sh
VITE_DEV_SERVER_KEY=/path/to/key
VITE_DEV_SERVER_CERT=/path/to/cert
```

## CLI output

We now output on the CLI when Vite certificates are in use. This will help users know when the Valet certificates have been auto-detected, etc.

Screen Shot 2022-12-09 at 12 30 28 pm

The `valetTls` option now has the following options.

```js
laravel({
  // ...

  // Auto-detection.
  // - Host is guessed via convention.
  // - Missing certificates are ignored.
  // _

  // Explicitly use Valet certificates.
  // - Host is guessed via convention.
  // - Errors are thrown when missing certificates. 
  valetTls: true,

  // Explicitly use Valet certificates.
  // - Host specified is used.
  // - Errors are thrown when missing certificates. 
  valetTls: 'tim.dev',

  // Opt-out of auto-detection.
  valetTls: false,
}),
```

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -544,11 +544,15 @@ function resolveDevelopmentEnvironmentServerConfig(host: string|boolean|null): {
     const certPath = path.resolve(configPath, 'Certificates', `${resolvedHost}.crt`)
 
     if (! fs.existsSync(keyPath) || ! fs.existsSync(certPath)) {
-        if (host !== null) {
-            throw Error(`Unable to find certificate files for your host [${resolvedHost}] in the [${configPath}/Certificates] directory. Ensure you have secured the site via the Herd UI or run \`valet secure\`.`)
+        if (host === null) {
+            return
         }
 
-        return
+        if (configPath === herdConfigPath()) {
+            throw Error(`Unable to find certificate files for your host [${resolvedHost}] in the [${configPath}/Certificates] directory. Ensure you have secured the site via the Herd UI.`)
+        } else {
+            throw Error(`Unable to find certificate files for your host [${resolvedHost}] in the [${configPath}/Certificates] directory. Ensure you have secured the site by running \`valet secure\`.`)
+        }
     }
 
     return {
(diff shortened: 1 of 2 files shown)


innocenzi:
> It just happens that it should match when configured correctly.

I understand the concern and that technically, the semantics don't match — however, in practice, this is the perfect fit for this option, I would argue. 

> If that doesn't work out, I'm not completely opposed to using APP_URL as long as we also check the working directory.

I think this would be the perfect compromise, if the solution above doesn't work.

innocenzi:
@timacdonald it's been a while, with the support or Herd now I thought this would be a good time to follow-up?

timacdonald:
@jessarcher and I have started talking about this again. We will circle back to it shortly and get things moving, whichever way.

timacdonald:
Will revisit this PR tomorrow and see if we can get it into shape for the `1.x` release.

PR #182, opened 2022-12-12
Update release docs

jessarcher:
This PR updates `RELEASE.md` to avoid an issue where the plugin could be built without the latest dependencies.

We don't commit the `package-lock.json` to this repository, so it's necessary to at least run `npm update` instead of an `npm install`. I figured it's probably safer to just start from a clean slate rather than updating, though.


driesvints (approved):
Cool, thanks @jessarcher! I want @taylorotwell to merge this one so he sees it and knows about it too.

PR #183, opened 2022-12-13
[0.7.x] Prevent Vite from injecting duplicate modulepreload tags

jessarcher:
The `@vite` directive in Laravel outputs `modulepreload` tags as of https://github.com/laravel/framework/pull/44096.

Vite itself also injects `modulepreload` tags into the `<head>` of the page via the JavaScript entrypoint, resulting in duplicates.

This PR aims to solve this by disabling the Vite feature in favour of the tags being injected by Laravel, which can include them in the initial HTML rather than having to wait until the entrypoint has been downloaded.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -130,6 +130,7 @@ function resolveLaravelPlugin(pluginConfig: Required<PluginConfig>): LaravelPlug
                         input: userConfig.build?.rollupOptions?.input ?? resolveInput(pluginConfig, ssr)
                     },
                     assetsInlineLimit: userConfig.build?.assetsInlineLimit ?? 0,
+                    modulePreload: userConfig.build?.modulePreload ?? false,
                 },
                 server: {
                     origin: userConfig.server?.origin ?? '__laravel_vite_placeholder__',


timacdonald:
I agree that the duplicate `modulepreload` tags on load isn't _great_, but I don't think we want to disable this. If we turn this off we introduce another step in the waterfall on all _subsequent_ navigations.

In a fresh Laravel Breeze installation, navigating from the "Welcome" page to the "Login" page has the following waterfall...

### Now

Screen Shot 2022-12-23 at 10 09 47 am


### Disabled preloading

Screen Shot 2022-12-23 at 10 07 48 am

Note: ignore the lengths of these as the scale is not the same. You should see there is a new step though.

Without preloading, on each subsequent navigation the browser has to download a JS file before it then knows what it needs. This waterfall could also explode with steps on a more complex system, as each file could have it's own imports, which are then needed.

Thus, I don't think disabling this feature is a good idea.

However, I agree that duplicate tags being. The solution to this is to ensure that Vite builds with the correct `ASSET_URL` in place.

If the `ASSET_URL` matches across Vite's built assets and Laravel, Vite won't output duplicate tags on first load.

jessarcher:
Thanks @timacdonald, that makes sense!

Issue #197, opened 2023-01-31
Issues with the hotfile and the host configuration


Madscientiste:
Currently the hot file will only store *one* interface address 

eg: `vite --host 0.0.0.0` will output `http://0.0.0.0:7894`, However you cannot connect to 0.0.0.0 directly.

Let say you have a system that has 2 network interfaces like 10.5.0.5 and 172.26.80.1, so then you can connect to either of these externally, and 127.0.0.1 internally

So having `http://0.0.0.0:7894` is not valid. 

what i've done to fix this is (as a workaround for now), i just put all my network addresses ( that vite already gives me ) in the hotfile
eg: `vite --host 0.0.0.0` will give me 
```
➜  Local:	http://localhost:7894
➜  Network: http://192.168.1.182:7894
➜  Network: http://192.168.137.1:7894
➜  Network: http://172.26.80.1:7894
``` 

and tell laravel to use whatever host is being used.
 
if i go to my laravel url (`http://172.26.80.1:8000`), it will give me 
```html
<script type="module" src="http://172.26.80.1:7894/@vite/client" />
<script type="module" src="http://172.26.80.1:7894/resources/js/app.jsx" />
```

There are other issues like it outputs the ipv6 loopback `http://[::1]:7894` when no --host is given 
which also doesn't work when fetching the assets.

And also where the given `base` isn't appended at the end, eg: when i set my `vite.config` with `base: '/something/'` 
i expect also to have `http://172.26.80.1:7894/something` in my hotfile 

But i think there are open or closed issues regarding these 2 ... 

I'm not sure if i was clear enough with my explanation, i hope it make sense 😅 
and please do tell me if there is anything i can help out with 👍 

### My configuration
- Laravel Vite Plugin Version: 0.7.3
- Laravel Version: 9.48.0
- Node Version: LTS-18.12.1
- NPM Version: 8.19.2
- Host operating system: Windows 11
- Running in Sail / Docker: No

<br />

jessarcher:
The issue of not being able to connect to `0.0.0.0` is a known issue with Windows. On Mac and Linux it behaves similarly to the loopback address when connecting to it. I'm not aware of a great solution for Windows yet - it's hard to simultaneously tell Vite to listen on all addresses but then only be able to use one of those addresses for asset links.

What should an ipv6 loopback address look like on Windows if not what is currently provided?

If you need to access it via ipv4 loopback, does it work if you specify the host as either `localhost` or `127.0.0.1`? 

[...]

PR #207, opened 2023-03-14
[0.8.x] Fix base

Madscientiste:
Fixes what i have mentionned in #197

> when i set my vite.config with base: '/something/'
i expect also to have http://172.26.80.1:7894/something in my hotfile
...
> If you specify a base path to vite, it means all the assets will be served from that base path. so if you have a base path of /somebasepath then all the assets will be served from http://localhost:3000/somebasepath/asset.js instead of http://localhost:3000/asset.js. This is useful if you want to serve vite from a subpath of your domain.

If i set a basepath and goes to the url without the base path it returns this:

```html
The server is configured with a public base URL of /admin/ - 
did you mean to visit <a href="[/admin/@vite/client](view-source:http://localhost:7894/admin/@vite/client)">/admin/@vite/client</a> instead?
```

for some reason it doesn't stop my assets to load, but it breaks HMR


jessarcher:
Sorry for the delay in getting to this one. It's a hard one to get my head around and make sure it doesn't introduce a breaking change, given all the different configurations we're trying to support and how `base` is set differently in dev mode vs build mode.

We recently merged #255, which hasn't been tagged yet but I think will allow you to solve your scenario. Feel free to re-open if it's not the case.

timacdonald:
Related: https://github.com/laravel/vite-plugin/pull/290

PR #217, opened 2023-05-09
[0.7.x] Polyfill import.meta.url for CJS builds

timacdonald:
`__dirname` is not supported in ES Modules. This PR migrates to using `import.meta.url` and polyfiling the CJS build with the more modern approach.

Read more about ES Modules and `__dirname` here: https://blog.logrocket.com/alternatives-dirname-node-js-es-modules/

The polyfill was taken from: https://github.com/evanw/esbuild/issues/1633


jessarcher (approved):

innocenzi:
Just FYI, dedicated build tools like https://github.com/unjs/unbuild provide this feature out-of-the-box. 👍 

I highly recommend using it instead of esbuild directly.

PR #221, opened 2023-05-11
[0.7.x] Add providers to default refresh paths

zepfietje:
Providers often contain some configuration which affect the UI as well.
Adding them to the default refresh paths makes for a smoother developer experience.


zepfietje:
Moved to https://github.com/filamentphp/filament/pull/6485.

PR #232, opened 2023-07-05
fix: support build in subfolders

devhau:
fix error: support build in subfolders
ex:
const vite = require("vite");
const pathSys = require("path");

function buildVite(root) {
  console.log(pathSys.join(root, "vite.config.js"));
  vite.build({
    root,
    configFile: pathSys.join(root, "vite.config.js"),
  });
}
buildVite('plugin/plugin1')
buildVite('plugin/plugin2')
buildVite('plugin/plugin2')
buildVite('plugin/plugin4')

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-    "name": "laravel-vite-plugin2",
+    "name": "laravel-vite-plugin",
     "version": "0.7.8",
     "description": "Laravel plugin for Vite.",
     "keywords": [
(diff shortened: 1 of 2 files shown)


jessarcher:
I'm not sure I understand the problem or why this change is necessary.

Could you perhaps provide minimal reproduction steps or a repository demonstrating the issue that this PR will solve?

devhau:
Hi @jessarcher 
You can see in my project:https://github.com/BitSudo/bitsudo-platform/blob/main/rebuild.js
when i use the current code, the folder contains the file vite.config.js then it makes the root directory.
when i run "node rebuild.js", it doesn't execute build code in example/resources folder.

jessarcher:
Hi @devhau,

I don't fully understand your use case or project, but it seems outside the scope of this plugin. The plugin was built to simplify the configuration of Vite in a typical Laravel application. It wasn't intended to be used in packages. In those situations, I configure Vite manually without using the Laravel plugin.

Note that the Laravel plugin aims to respect any Vite/Rollup options you have specified in your `vite.config.js`, so if you need to configure `build.rollupOptions.input`, you may do this directly in your configuration.

E.g.

```js
export default defineConfig({
    plugins: [
        // ...
    ],

    build: {
        rollupOptions: {
            input: // Your inputs here.
        },
    },
});

PR #233, opened 2023-07-20
Support Laravel Herd

claudiodekker:
This PR adds Laravel Herd support, with fallback support to Valet.

The reason for prefering Herd over Valet, has to do with that Valet users are likely to upgrade to Herd, but Herd users aren't as likely to downgrade back to Valet. Simultaneously, when Herd is uninstalled, removing the folder that's being looked up here is part of the required steps, as per https://herd.laravel.com/docs/1/troubleshooting/uninstalling

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -55,7 +55,7 @@ interface PluginConfig {
     refresh?: boolean|string|string[]|RefreshConfig|RefreshConfig[]
 
     /**
-     * Utilise the valet TLS certificates.
+     * Utilise the Herd or Valet TLS certificates.
      *
      * @default false
      */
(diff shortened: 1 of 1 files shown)


driesvints (approved):

src/index.ts
```diff
@@ -524,17 +526,30 @@ function resolveValetServerConfig(host: string|boolean): {
...
+
+    if (fs.existsSync(herdConfigPath)) {
+        return herdConfigPath
+    }
+
+    return path.resolve(os.homedir(), '.config', 'valet');
+}
+
 /**
  * Resolve the valet valet host for the current directory.
```
claudiodekker, line 543:
```suggestion
 * Resolve the Valet host for the current directory.
```

timacdonald:
Thanks, @claudiodekker. 

I have also deprecated the `valetTls` config in favour of the more generic `detectTls` config option.

Updated docs: https://github.com/laravel/docs/pull/8925

PR #241, opened 2023-08-29
[0.8] Fix issue with `0.0.0.0` network resolution

timacdonald:
This PR only changes functionality when running inside of Laravel Sail.

When determining the URL to place in the hot file we currently fallback to the resolved configuration's `server.host`. The plugin sets the resolved configuration's `server.host` to `0.0.0.0` when running inside of Laravel Sail if no user configured value is present.

This means that `0.0.0.0` will end up in the hot file, by default, when running the Vite server inside of Laravel Sail.

This address is fine within the container and is where the Vite server should listen, however communicating from outside the container via this host can be problematic. Safari on Mac is unable to communicate with this host, while Chrome and Firefox are okay. Additionally, communicating with this host on Windows is a problem across the board. We had to add [special documentation to the Laravel Docs](https://laravel.com/docs/10.x/vite#configuring-hmr-in-sail-on-wsl2) just for Windows to work around this issue.

This change should fix the issue in Safari on Mac and hopefully across the board on Windows as well. We can potentially remove the special docs as well.

Now, when we detect that the Vite server is running within Laravel Sail, we will fallback to `localhost` instead of `0.0.0.0` for the value that is written to the hot file. The Vite server will still listen on `0.0.0.0`. Any user specified configuration will take precedence.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -107,6 +107,7 @@ export default function laravel(config: string|string[]|PluginConfig): [LaravelP
 function resolveLaravelPlugin(pluginConfig: Required<PluginConfig>): LaravelPlugin {
     let viteDevServerUrl: DevServerUrl
     let resolvedConfig: ResolvedConfig
+    let userConfig: UserConfig
 
     const defaultAliases: Record<string, string> = {
         '@': '/resources/js',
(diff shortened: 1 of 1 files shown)


jessarcher (approved):

TheBlckbird:
I’m currently not at home, but I’m pretty sure I had this issue ouside of Sail as well. Can you wait with the merging until Saturday?

TheBlckbird:
nvm it works outside of sail, @timacdonald

timacdonald:
Thanks for testing, @TheBlckbird

TheBlckbird:
thanks :)

PR #246, opened 2023-09-16
Upgrade vitest

sapphi-red:
This PR upgrades Vitest from ^0.25.2 to ^0.34.4 (latest).

### Reasons

The ecosystem-ci failed with https://github.com/vitejs/vite/pull/13944.
I'm not sure what the actual reason is, but it passed locally if I upgraded Vitest. Maybe Vitest was relying on CJS ssr feature in old versions.
I guess there won't be any problem with upgrading it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,7 @@
         "eslint": "^8.14.0",
         "typescript": "^4.6.4",
         "vite": "^4.0.0",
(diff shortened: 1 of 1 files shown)


timacdonald (approved):
Thanks so much for sending this over!

PR #251, opened 2023-10-07
[1.x] Introduce `clean-orphaned-assets` binary

timacdonald:
When deploying a site and building your assets during a deploy, if you have enough traffic, you can run into issues while building vite assets.

Vite will remove the `manifest.json` and all the existing assets in `public/build/assets/*` before it writes the new manifest and build assets.

This means for a brief time there is no manifest and also in-flight responses will reference assets that no longer exist on the server.

To resolve this you can build your assets and tell Vite not to clean the build directory:

```sh
npx vite build --emptyOutDir=false
```

Vite will now write over the manifest and will not remove the existing assets.

This is great, and generally good practice, but it introduces another problem: nothing is cleaning up the old assets. They will continue to grow and overtime these will build up to be a decent size.

This is where the new command comes in. It will read the current manifest and clean out any old build assets that are no longer referenced in the current manifest.

When running this command it is recommend you give a little time after building the new assets before running the command. You need to consider how long a response referencing old assets might take to return to the user and for them to then request the old assets.

```
npx vite build --emptyOutDir=false

# wait n seconds before cleaning out the old assets
sleep 3

npx clean-orphaned-assets
```

## General Usage

Clean orphaned assets for non-SSR builds using conventional paths for the manifest and assets.

```
npx clean-orphaned-assets
```

**Of course I would love this to not be so verbose, but it needs to not conflict with other scripts that may have been configured.

Clean orphaned assets for SSR builds using conventional paths for the manifest and assets.

```
npx clean-orphaned-assets --ssr
```

### Flags & Options

`--ssr`: Clean SSR assets.
`--quiet`: No output written.
`--dry-run`: Does not actually remove the files from the directory.
`--manifest={path}`: Specify a custom path to the manifest file.
`--assets={path}`: Specify a custom path to the assets directory.

## Notes

- I just let the "missing file" errors fall through from Node to keep things simple.


.eslintrc.cjs
timacdonald, line 1: This was required to make the package `type: module`. [ESLint doesn't support ES modules](https://eslint.org/docs/latest/use/configure/configuration-files#configuration-file-formats).

package.json
```diff
@@ -1,5 +1,6 @@
 {
     "name": "laravel-vite-plugin",
+    "type": "module",
```
timacdonald, line 3: This is required to use MJS in the purge script.

innocenzi: You can also simply write the binary in TypeScript and use a build tool to transpile it back to CJS.

- [`unbuild`](https://github.com/unjs/unbuild) is great, I use it in a lot of projects. It's built on top of Rollup.
- [`tsup`](https://github.com/egoist/tsup) is also a good alternative. It's built on top of esbuild.

I personally recommend `unbuild` and I can help you setting that up if needed 👍

timacdonald: We'll be making this change as part of the `1.x` release anyway, so I think it's good with this.

timacdonald:
See: https://github.com/laravel/framework/issues/48964

timacdonald:
Will review this tomorrow and get it ready to be reviewed / merged again

francoism90:
@timacdonald Would you recommend running `npx clean-orphaned-assets` for production/on deployment?

Thanks!

timacdonald:
@francoism90 I would. You should use `--dry-run` first to make sure it is going to delete what you expect. Will have docs shortly.

It is only really useful if you are also using the `--emptyOutDir=false` when building your assets.

Issue #253, opened 2023-10-09
Vite 5.0 Support


gabrielrbarbosa:
```
$ npm i
npm ERR! code ERESOLVE
npm ERR! ERESOLVE could not resolve
npm ERR!
npm ERR! While resolving: laravel-vite-plugin@0.8.0
npm ERR! Found: vite@5.0.0-beta.4
npm ERR! node_modules/vite
npm ERR!   dev vite@"^5.0.0-beta.4" from the root project
```

driesvints:
No stable v5 has been released yet. I'm sure @jessarcher and @timacdonald are on this as soon this hits stable.

driesvints:
Not yet. We've been too busy with other projects like Laravel Pulse. You'll see an update as soon as we have something to share.

PR #254, opened 2023-10-22
Handle Vite 5 types changes

bluwy:
Vite 5 removes the `https: boolean` type and as it doesn't have much use by default. A plugin or manual configuration is always needed for `https` to actually work (e.g. certs). (https://github.com/vitejs/vite/pull/14681)

This PR updates the types breaking change caused by it. It uses `@ts-ignore` so it works on both Vite 4 and 5. The `as any` is also needed so the `config` hook doesn't error by TS due to incorrect type returned by the function.

---

### Extra notes unrelated to PR

It seems like before (Vite 4), users could do `vite --https` or `https: true` in `vite.config.js` to get the right certs generate by `laravel-vite-plugin`. In Vite 5, this isn't possible anymore except if `https: {}` in `vite.config.js`. Though it isn't quite exactly documented in the [laravel docs](https://laravel.com/docs/10.x/vite#working-with-a-secure-development-server).

Just want to note this in case you want to handle this differently in the future. For example, the `detectTls` could be used to enable `https` internally. Right now it doesn't due to the existing logic here:

https://github.com/laravel/vite-plugin/blob/7abca675b82dd8a4fd23470f978110c3d083a0a1/src/index.ts#L163-L166

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -160,9 +160,11 @@ function resolveLaravelPlugin(pluginConfig: Required<PluginConfig>): LaravelPlug
                             ...serverConfig.hmr,
                             ...(userConfig.server?.hmr === true ? {} : userConfig.server?.hmr),
                         },
+                        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
                         // @ts-ignore `https: boolean` is removed in Vite 5
-                        https: userConfig.server?.https === false ? false as any : {
+                        https: userConfig.server?.https === false ? false as never : {
                             ...serverConfig.https,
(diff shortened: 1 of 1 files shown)


gabrielrbarbosa:
#253

timacdonald:
Really appreciate you sliding into our Laravel plugin, @bluwy, to help get us ready for Vite 5 ❤️

I've take the changes here and added them to our single PR to upgrade to Vite 5. I'll close this one off, but thank you so much.

We are also gonna drop Vite 4 support, so it should make things a little smoother with type changes.

bluwy:
No problem! 👍

PR #258, opened 2023-11-04
Fix types for moduleResolution=nodenext/bundler

sapphi-red:
The types condition was wrongly declared.
https://arethetypeswrong.github.io/?p=laravel-vite-plugin%400.8.1

This PR fixes that by

- Emitting both types for require and import for `.` export
  - A single type file cannot be used for both ESM and CJS. See https://github.com/arethetypeswrong/arethetypeswrong.github.io/blob/main/docs/problems/FalseCJS.md for more details.
- Removing types condition
  - `types` condition must be the first property. Instead of changing the order, I simply removed the condition because TypeScript will also look up the adjacent type file. See https://github.com/arethetypeswrong/arethetypeswrong.github.io/blob/main/docs/problems/FallbackCondition.md for more details.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,6 @@
         },
         "./inertia-helpers": {
             "import": "./inertia-helpers/index.js",
-            "types": "./inertia-helpers/index.d.ts",
             "node": "./inertia-helpers/index.js"
         }
     },
--- a/package.json
+++ b/package.json
@@ -17,8 +17,7 @@
     "exports": {
         ".": {
             "import": "./dist/index.mjs",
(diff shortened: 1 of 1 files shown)


timacdonald:
Thank you so much for dropping by our Laravel world to help us out with the upgrade, @sapphi-red.

We've got a working branch where we have already made the type module changes, thankfully.

I'm going to close this one as we I'm creating a single PR to house all our Vite 5 upgrade work to make it easier for review etc.

Thanks again. We really appreciate the Vite team ❤️

PR #260, opened 2023-11-15
[0.8.x] allow disabling default prompt

innocenzi:
This pull request allows disabling the default prompt that tells the Laravel version, plugin version and `APP_URL`.

My situation is that I am using a wrapper around this plugin that displays its own information, but normal users could want to disable it too.

Happy to rename the option if needed :)

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -73,6 +73,13 @@ interface PluginConfig {
      * Transform the code while serving.
      */
     transformOnServe?: (code: string, url: DevServerUrl) => string,
+
+    /**
+     * Show a Laravel-specific prompt when starting the development server.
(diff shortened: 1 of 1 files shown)


timacdonald:
@innocenzi I'm personally not super keen to add more config.

Can you use Vite's [Custom logger](https://vitejs.dev/config/shared-options.html#customlogger) to achieve this?

innocenzi:
Hey @timacdonald, thanks for at least considering. I'm going to use my own plugin implementation instead of wrapping to ease things up on my side instead. Sorry for bothering 👍

timacdonald:
No bother at all ✌️

PR #263, opened 2023-11-17
Allow vite 5

ziming:
Allow vite 5 in package.json


driesvints:
This seems to break the test suite. We're working on this and hope to have a Vite 5 compatible version soon.

PR #269, opened 2023-12-13
[1.x] Vite 5

timacdonald:

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,5 +1,32 @@
 # Upgrade Guide
 
+## `0.x` to `1.x`
+
+This version adds support for Vite 5 while removing support for Vite 3 and 4.
+
+### Changes
+
+1. Dropped support for CJS: https://github.com/laravel/vite-plugin/pull/235
+2. Dropped support for a `boolean` value of `server.https` option in `vite.config.js`: https://github.com/laravel/vite-plugin/pull/254
+
+### Upgrade path
+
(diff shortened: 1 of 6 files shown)


UPGRADE.md
```diff
@@ -1,5 +1,36 @@
...
+
+The following command will update `laravel-vite-plugin` and `vite` only:
+
+```
+npm install vite@^5.0.0 laravel-vite-plugin@^1.x
+```
+
+You should also check the upgrading guide and changelogs for any packages you update. The [`vite` migrations guide](https://vitejs.dev/guide/migration.html) is available on their website.
+
+2. Ensure your `package.json` contains `"type": "modules"`.
```
shoeyn, line 24: Should be `"type": "module"` instead of "modules"

timacdonald: Thanks!

AntonOfTheWoods:
Any visibility on when the `1.x` branch might be released?

src/index.ts
```diff
@@ -160,7 +160,7 @@ function resolveLaravelPlugin(pluginConfig: Required<PluginConfig>): LaravelPlug
                             ...serverConfig.hmr,
                             ...(userConfig.server?.hmr === true ? {} : userConfig.server?.hmr),
                         },
-                        https: userConfig.server?.https === false ? false : {
+                        https: typeof userConfig.server?.https !== 'undefined' ? userConfig.server.https : {
                             ...serverConfig.https,
                             ...(userConfig.server?.https === true ? {} : userConfig.server?.https),
```
Jubeki, line 165: @timacdonald I took a look at the PR and found something which doesn't make sense to me without diving much deeper into the source:

The upgrade guide says that boolean support for `https` is dropped, but there is still a ```=== true``` check here. Is this intended? (I only want to make you aware, I don't know if this is needed or not)

timacdonald: Thanks! Fixed that one up.

No idea why typescript isn't complaining about that 😵

PR #271, opened 2023-12-15
[0.8.x] Fallback pages

timacdonald:
This PR is an alternative to https://github.com/laravel/vite-plugin/pull/252 and https://github.com/laravel/vite-plugin/pull/259.

It allows the `path` to be a string (current behaviour) or an array (new behaviour).

This enables two features:

1. A "fallback" so that users do not see a "whitescreen of death", where `./Pages/Whoops.vue` is the fallback:

```js
createInertiaApp({
    // ...
    resolve: (name) => resolvePageComponent([`./Pages/${name}.vue`, './Pages/Whoops.vue'], import.meta.glob('./Pages/**/*.vue')),
});
```

Of course, if the fallback isn't present things can still go wrong.

2. Easier migration from one file type to another without having to deal with error control, e.g., when migrating from `.jsx` to `.tsx`

```js
createInertiaApp({
    // ...
    resolve: (name) => resolvePageComponent([`./Pages/${name}.tsx`, `./Pages/${name}.jsx`], import.meta.glob('./Pages/**/*.[tj]sx')),
});
```

### DIFF
--- a/src/inertia-helpers/index.ts
+++ b/src/inertia-helpers/index.ts
@@ -1,9 +1,13 @@
-export async function resolvePageComponent<T>(path: string, pages: Record<string, Promise<T> | (() => Promise<T>)>): Promise<T> {
-    const page = pages[path]
+export async function resolvePageComponent<T>(path: string|string[], pages: Record<string, Promise<T> | (() => Promise<T>)>): Promise<T> {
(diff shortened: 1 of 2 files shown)


sh-sabbir:
@timacdonald Now that I see yours, it's indeed a better solution than mine

PR #273, opened 2023-12-19
[1.x] Simpler conditional

Jubeki:
`userConfig.server?.https` can either be undefined or be a user-defined object.

Currently there is an unnecessary conditional, because in the first ternary condition it is already checked if this configuration is undefined or not, therefor making the second ternary condition unnecessary. This PR solves this issue.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -162,7 +162,7 @@ function resolveLaravelPlugin(pluginConfig: Required<PluginConfig>): LaravelPlug
                             ...serverConfig.hmr,
                             ...(userConfig.server?.hmr === true ? {} : userConfig.server?.hmr),
                         },
-                        https: typeof userConfig.server?.https !== 'undefined' ? userConfig.server.https : serverConfig.https,
+                        https: userConfig.server?.https ?? serverConfig.https,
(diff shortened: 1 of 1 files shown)


timacdonald (approved):

src/index.ts
```diff
@@ -162,10 +162,7 @@ function resolveLaravelPlugin(pluginConfig: Required<PluginConfig>): LaravelPlug
                             ...serverConfig.hmr,
                             ...(userConfig.server?.hmr === true ? {} : userConfig.server?.hmr),
                         },
-                        https: typeof userConfig.server?.https !== 'undefined' ? userConfig.server.https : {
-                            ...serverConfig.https,
-                            ...(typeof userConfig.server?.https === 'undefined' ? {} : userConfig.server?.https),
-                        },
+                        https: typeof userConfig.server?.https !== 'undefined' ? userConfig.server.https : serverConfig.https,
```
GrahamCampbell: Can't this be simplified to just

```
https: userConfig.server?.https ?? serverConfig.https,
```

GrahamCampbell (changes requested):

PR #285, opened 2024-02-19
[1.x] Fix HMR issue when `resources/lang` directory doesn't exist and a symlink is present in the root directory

### LINKED ISSUE #284: No HMR in weird symlink scenario
### Vite Plugin Version

1.0.1

### Laravel Version

10.44.0

### Node Version

21.6.2

### NPM Version

10.2.4

### Operating System

Linux

### OS Version

Ubuntu 20.04.6

### Web browser and version

Safari Version 17.3 (19617.2.4.11.8)

### Running in Sail?

No

### Description

Certain symlinks break hmr file changes from being recognised. So far, I have only noticed `pint.json` but I presume it's other similar type config that Vite might recognise?

The weird thing here is that if I explicitly pass to the `refresh` option [the `refreshPaths` array](https://github.com/laravel/vite-plugin/blob/f2f0918a4cc982c91280afdf17ed9108c1fac89f/src/index.ts#L91) instead of `true`, but I remove `"resources/lang/**"` from that list, it works again.

Another possibly relevant point is that if I start `npm run dev` with a valid configuration/set of file then change them while it's running, the hmr file changes are still detected, so must be something at initial run.

FYI, the reason we had a `pint.json` symlink in the first place was to share Pint configuration across multiple projects.

### Steps To Reproduce

https://github.com/legecha/vite-plugin-bug-report

See [commits](https://github.com/legecha/vite-plugin-bug-report/commits/main/) for examples of working and non-working states.

jessarcher:
This PR works around an obscure issue of HMR breaking when a symlink is present in the root directory along with one of the configured refresh paths not existing but sharing a common parent with a file being monitored.

While obscure, half of this combination can happen easily because we configure the `resources/lang` directory to be watched by default, but this no longer exists in the Laravel skeleton. All a user then needs to do to trigger the issue is to create any symlink in the project's root directory.

I couldn't track down exactly why this issue occurs, but it seems sensible to restrict the default refresh paths to only those that exist.

Fixes #284.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -95,7 +95,7 @@ export const refreshPaths = [
     'resources/lang/**',
     'resources/views/**',
     'routes/**',
-]
+].filter(path => fs.existsSync(path.replace(/\*\*$/, '')))
 
 /**
  * Laravel plugin for Vite.
--- a/tests/index.test.ts
+++ b/tests/index.test.ts
@@ -2,6 +2,23 @@ import { afterEach, describe, expect, it, vi } from 'vitest'
 import laravel from '../src'
 import { resolvePageComponent } from '../src/inertia-helpers';
 
+vi.mock('fs', async () => {
+    const actual = await vi.importActual<typeof import('fs')>('fs')
+
+    return {
+        default: {
+            ...actual,
+            existsSync: (path: string) => [
+                'app/Livewire/',
+                'app/View/Components/',
+                'resources/views/',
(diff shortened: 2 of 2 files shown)


timacdonald (commented):
Weirdest. Bug. Ever.

This bug is essentially:

> If you live in a council flat...next to a river...but are not blind...
> **Bernard Black**. Black Books.

PR #290, opened 2024-03-18
Append base to hot file

timacdonald:
fixes #288

Fixes an issue where the `base` path is not included in the `/public/hot` file. This issue only impacts the `npm run dev` command. When building, assets are appended with the configured base.

## `base` description

Screenshot 2024-03-18 at 13 04 32

## What

If we were to set the `base` path to `/frontend/`:

```js
import { defineConfig } from 'vite';
import laravel from 'laravel-vite-plugin';

export default defineConfig({
    base: '/frontend/',
    plugins: [
        laravel({
            input: ['resources/css/app.css', 'resources/js/app.js'],
            refresh: true,
        }),
    ],
});
```

Then the dev server will be appended with `/frontend/`:

Screenshot 2024-03-18 at 13 08 15

However, the hot file we write will not contain the `/frontend` base path:

Screenshot 2024-03-18 at 13 08 43

So we do not create the correct URLs in Laravel. With this change in place, the hot file now contains the correct base.

Screenshot 2024-03-18 at 13 12 26

## Implementation notes

Vite normalizes the `base` value before we get access to it. The following table shows how Vite normalizes different values and what we end up appending to the hot file.

The only valid values for this option are `undefined`, an empty string, `./`, or an absolute URL, i.e., `https://example.com/...` or `/...`

| User config | Normalized | Hot file
| --- | --- | --- |
| `undefined` | `"/"` | `""` |
| `""` | `"/"` | `""` |
| `"/"` | `"/"` | `""` |
| `"./"` | `"/"` | `""` |
| `"/frontend"` | `"/frontend/"` | `"/frontend"` |
| `"/frontend/"` | `"/frontend/"` | `"/frontend"` |
| `"https://example.com"` | `"/"` | `""` |
| `"https://example.com/"` | `"/"` | `""` |
| `"https://example.com/frontend"` | `"/frontend/"` | `"/frontend"` |
| `"https://example.com/frontend/"` | `"/frontend/"` | `"/frontend"` |

PR #296, opened 2024-05-16
Include base in hotFile without modifying server.origin replacement

danielztolnai:
Unfortunately #290 - while fixing the `hotFile` problem - broke the transform function's origin replacement when using a base in the configuration resulting in broken asset URLs.

Vite generates URLs using the `origin` and the `base`. As this plugin now replaces the `origin` (`__laravel_vite_placeholder__`) by default with the `origin+base`, the base is included twice in the asset URLs, which results in the given assets not loading.

On the other hand, the original problem that #290 solved was real as well, as Laravel did not take the `base` into account before this PR and the main files did not load when setting `base` in the configuration. This means that the `base` has to be included in the `hotFile`. It should also be included when setting the `origin` explicitly, which is a special case that #290 did not handle.

This pull request includes the `base` in the `hotFile`, but leaves the origin (or more precisely the replacement) as is. This should both solve the original problem and avoid including the `base` in the asset URLs twice.

|                    | before                                                                                     | after                                                                                      |
|--------------------|--------------------------------------------------------------------------------------------|--------------------------------------------------------------------------------------------|
| `public/hot` content | `https://test.localhost:443/vite-test-base`                                                  | `https://test.localhost:443/vite-test-base`                                                  |
| asset URL example  | `https://test.localhost/vite-test-base/vite-test-base/resources/images/test.svg`   | `https://test.localhost/vite-test-base/resources/images/test.svg`                  |
| chunk URL example  | `https://test.localhost/vite-test-base/node_modules/.vite/deps/chunk-NDQ3I6BD.js?v=df756ebe` | `https://test.localhost/vite-test-base/node_modules/.vite/deps/chunk-NDQ3I6BD.js?v=df756ebe` |

This change should not be a breaking change as it is a fix to something that changed in vite 5 compared to vite 4 as mentioned [here](https://github.com/laravel/vite-plugin/pull/290#issuecomment-2002774467).

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -84,7 +84,7 @@ interface LaravelPlugin extends Plugin {
     config: (config: UserConfig, env: ConfigEnv) => UserConfig
 }
 
-type DevServerUrl = `${'http'|'https'}://${string}:${number}${string}`
+type DevServerUrl = `${'http'|'https'}://${string}:${number}`
 
 let exitHandlersBound = false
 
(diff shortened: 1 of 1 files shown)


timacdonald (approved):
Thanks!

jessarcher (approved):

driesvints:
Thanks! We'll try to get this one out as soon as we're able.

driesvints:
Released!

PR #307, opened 2024-11-04
Look for certificates in valet linux config directory

jameshulse:
Hi, this pull request adds another search path to support the [valet-linux](https://github.com/cpriego/valet-linux) project.

I understand that this project isn't officially supported, but a small change in this Vite plugin would allow our team to greatly simplify our Vite configuration and support developers using Linux or Mac with Herd.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -218,7 +218,7 @@ function resolveLaravelPlugin(pluginConfig: Required<PluginConfig>): LaravelPlug
                                 server.config.logger.info(`  ${colors.green('➜')}  Using Herd certificate to secure Vite.`)
                             }
 
-                            if (resolvedConfig.server.https.key.startsWith(valetMacConfigPath())) {
+                            if (resolvedConfig.server.https.key.startsWith(valetMacConfigPath()) || resolvedConfig.server.https.key.startsWith(valetLinuxConfigPath())) {
                                 server.config.logger.info(`  ${colors.green('➜')}  Using Valet certificate to secure Vite.`)
                             }
                         }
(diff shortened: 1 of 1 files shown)


src/index.ts
```diff
@@ -218,7 +218,7 @@ function resolveLaravelPlugin(pluginConfig: Required<PluginConfig>): LaravelPlug
                                 server.config.logger.info(`  ${colors.green('➜')}  Using Herd certificate to secure Vite.`)
                             }
 
-                            if (resolvedConfig.server.https.key.startsWith(valetConfigPath())) {
+                            if (resolvedConfig.server.https.key.startsWith(valetMacConfigPath()) || resolvedConfig.server.https.key.startsWith(valetLinuxConfigPath())) {
                                 server.config.logger.info(`  ${colors.green('➜')}  Using Valet certificate to secure Vite.`)
                             }
```
timacdonald, line 223: @jameshulse, I've just made this change to ensure the Linux version also sees the log message in the Vite server output. Could you confirm that is working as expected?

timacdonald (approved):
These changes look good to me.

I've renamed the first-party valet function to `valetMac`. This is to make it easier to understand what is going on with both functions and also aligns with `herdMac` / `herdWindows`.

Although you could say that we want to keep the first-party method as just `valet`, it makes it easier to know there is another version to also check when working with the code.

src/index.ts
jameshulse: I sure do:

image

jameshulse:
Good shout. Cheers for that Tim.

Shameless plug: if the team are looking for more early Laravel Cloud adopters, we are keen to get off self hosting by the end of the year.

src/index.ts
timacdonald: Thanks!

taylorotwell:
Thanks @jameshulse!

Issue #320, opened 2025-02-12
package-lock.json / pnpm-lock.yml not checked in


hirasso:
### Vite Plugin Version

1.2.0

### Laravel Version

n/a

### Node Version

n/a

### NPM Version

n/a

### Operating System

macOS

### OS Version

n/a

### Web browser and version

n/a

### Running in Sail?

No

### Description

Hi there! I just wanted to source-dive laravel-vite-plugin. I'm using [ni](https://github.com/antfu-collective/ni) so that I always install packages using the preferred package manager of each project. Running that I noticed that there is no package-lock.json (or any other npm lockfile) present in laravel-vite-plugin. Was it forgotten to check that file into VCS or is this a conscious decision?

### Steps To Reproduce

```
git clone git@github.com:laravel/vite-plugin.git laravel-vite-plugin
cd laravel-vite-plugin
ni
```
Output of the `ni` command:
```shell
? Choose the agent › - Use arrow-keys. Return to submit.
❯   npm
    yarn
    pnpm
    bun
    deno
```
`ni` asks which type of package manager it should use. That's the default behavior if it can't find a lock file.

timacdonald:
As a package, this does not include a lock file.

You can just use the latest dependencies with whatever package manager you like.

PR #322, opened 2025-02-15
Ignore .DS_Store and common package manager lock files

hirasso:
- ignore .DS_Store
- ignore pnpm-lock.yaml
- ignore yarn.lock

As discussed in #320 , it is the opinion of laravel-vite-plugin to not check in package manager lock files. Also, no specific package manger should be enforced as explained by @timacdonald in the [issue comment](https://github.com/laravel/vite-plugin/issues/320#issuecomment-2655010429). 

If this is actually the intention, all common package manager lock files should be ignored in the `.gitignore` file, to prevent committing them by accident.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,8 @@
+.DS_Store
 /node_modules/
 /dist/
 /inertia-helpers/
 /npm-debug.log
 /package-lock.json
+/pnpm-lock.yaml
+/yarn.lock


timacdonald:
In that case I would recommend using `npm`. Feel we will get a stream of PRs adding lock files.

hirasso:
You mean you would then prefer to check in the package-lock.json? Should I create a PR for that?

timacdonald:
I don't think we should check-in the lock file.

I think you should use `npm`, instead of another package manager, when developing the `laravel-vite` plugin.

hirasso:
I do not understand this. It is so common to have at least one lock file checked in. And it is also very common to ignore DS_Store files.

timacdonald:
Re: lock files, for projects where specific versions are important it is good to commit a lock file. For packages, such as this one, I would say it is generally not common. I would guess that less than 10% of Laravel packages have lock files committed, as an example, and when they do it is because they provide a built output (e.g., CSS or JS).

For `.DS_Store`, these should be ignored in your global git ignore. `.DS_Store` files are not project specific and have nothing to do with the Laravel Vite plugin.

hirasso:
Interesting. Thank you for expanding on the topic! It's the first time I'm hearing about a _global_ git ignore file...

PR #325, opened 2025-04-14
Use rollup types from Vite

sapphi-red:
Since `rollup` is not declared as a dependency or a peerDependency, this import fails in some cases (see https://rushjs.io/pages/advanced/phantom_deps/ for more details).

That was causing the ecosystem-ci to fail.
https://github.com/vitejs/vite-ecosystem-ci/actions/runs/14437826501/job/40481878417#step:8:697

This PR fixes that by using the re-exported types from Vite.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -4,15 +4,14 @@ import os from 'os'
 import { fileURLToPath } from 'url'
 import path from 'path'
 import colors from 'picocolors'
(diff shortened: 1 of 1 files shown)


timacdonald (approved):
Appreciate it, @sapphi-red

sapphi-red:
@timacdonald Hi :wave: 
Would it be possible to get this PR merged? Right now, the ecosystem CI is failing because of this type error, which makes it hard to catch other potential issues.

PR #327, opened 2025-06-12
Valet/Herd Preference + Smarter Detection

jasoryeh:
Hi, oddly enough, I use Valet+PHPMonitor and Herd separately and at times swap between them, and discovered that once Herd is installed, the Vite plugin breaks when I try to use this plugin with pure valet sites since the plugin looks for Herd configs first and ignores valet after that.

To remedy this, I added an env `LARAVEL_PREFER_DEVELOPMENT_ENVIRONMENT` (which can be `herd` or `valet`) that will prefer one or the other if both exist.

I also added "smarter" detection for these certificates by checking for certificates of the current site in all available directories first instead of choosing Herd first just because its `valet` directory exists.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -543,6 +543,8 @@ function resolveDevelopmentEnvironmentServerConfig(host: string|boolean|null, pr
         } else {
             candidateConfigs.push(...configsAvailable.herd, ...configsAvailable.valet)
         }
+    } else {
+        candidateConfigs.push(...configsAvailable.herd, ...configsAvailable.valet)
     }
 
     const configPath = candidateConfigs[0]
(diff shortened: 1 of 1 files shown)


timacdonald:
Would it not be possible to use the existing `VITE_DEV_SERVER_KEY` and `VITE_DEV_SERVER_CERT` environment variables?

taylorotwell:
Thanks for your pull request to Laravel!

Unfortunately, I'm going to delay merging this code for now. To preserve our ability to adequately maintain the framework, we need to be very careful regarding the amount of code we include. 

If applicable, please consider releasing your code as a package so that the community can still take advantage of your contributions!

PR #338, opened 2026-01-05
fix: missing 2.0.0 release notes in changelog

james2doyle:
I went to update my dependencies and before I did, I checked the release notes for the latest version. I saw no breaking changes so I bumped my version and installed. Except it wouldn’t install. I got an error saying that Vite 7 is required. I checked the changelog and saw no such note about the update/change.

Turns out the notes for 2.0.0 were missing that described the change.


timacdonald:
Thanks!

PR #341, opened 2026-03-12
Vite 8 

JonPurvis:
Hey,

This PR updates the plugin to use Vite V8, which offers significant speed improvements on build times as it swaps usage of `Rollup` and `esbuild` to `rolldown` and `Oxc`.

From my own testing in one of my apps, I noticed a ~85% difference in build time:

| Version | Build Time |
|--------|--------|
| V7 | ~2s |
| V8 | ~300ms | 

I've removed packages from `package.json` that are now no longer needed and updated the nessassary scripts within `package.json` too. Tests have also been updated to use `rollDown` and `index.ts` has been updated to support both `rollUp` and `rollDown` as Vite V8 supports both.


NeilInnes:
Yes please 😀

entity:
+1 to these! Tried Vite v8 today and it seems to be a substantial difference.

JonPurvis:
Yeah, I compared build times yesterday for one of my apps:

- V7 built in 1.96s
- V8 built in **326ms**

entity:
> Yeah, I compared build times yesterday for one of my apps:
> 
> * V7 built in 1.96s
> * V8 built in **326ms**

Wow these are crazy! Mine dropped by ~50%
image

juliuskiekbusch:
Hey, also really looking forward to Vite 8 Support in the Laravel Vite Plugin!

Are the following packages even needed? I think Vite 8 uses Rolldown instead of Rollup + esbuild, so maybe we can remove these dependencies (have not further investigated yet, just a thought).

- esbuild
- eslint => oxlint (same toolchain as Vite): https://oxc.rs/docs/guide/usage/linter/migrate-from-eslint.html
    - @eslint/eslintrc
    - @eslint/js
    - @typescript-eslint/eslint-plugin
    - @typescript-eslint/parser
    
Just as said above, I didn't check this, I only shared my thoughts.

JonPurvis:
Good point @juliuskiekbusch 

I've pushed an update to this branch to remove dependencies that are no longer needed! Everything still working as intended, builds are still speedy:

Screenshot 2026-03-13 at 14 06 44

juliuskiekbusch:
Vitest can also be upraded from v3 to v4. 👍

package.json
```diff
@@ -35,28 +35,20 @@
...
-        "@typescript-eslint/eslint-plugin": "^8.0.0",
-        "@typescript-eslint/parser": "^8.0.0",
-        "esbuild": "0.25.6",
-        "eslint": "^9.0.0",
-        "globals": "^16.3.0",
+        "oxlint": "^1.0.0",
         "typescript": "^5.0.0",
-        "vite": "^7.0.0",
+        "vite": "^8.0.0",
         "vitest": "^3.0.0"
```
juliuskiekbusch:
```suggestion
        "vitest": "^4.1.0"
```

src/index.ts
```diff
@@ -148,8 +148,8 @@ function resolveLaravelPlugin(pluginConfig: Required<PluginConfig>): LaravelPlug
                     manifest: userConfig.build?.manifest ?? (ssr ? false : 'manifest.json'),
                     ssrManifest: userConfig.build?.ssrManifest ?? (ssr ? 'ssr-manifest.json' : false),
                     outDir: userConfig.build?.outDir ?? resolveOutDir(pluginConfig, ssr),
-                    rollupOptions: {
-                        input: userConfig.build?.rollupOptions?.input ?? resolveInput(pluginConfig, ssr)
+                    rolldownOptions: {
+                        input: userConfig.build?.rolldownOptions?.input ?? resolveInput(pluginConfig, ssr)
```
pascalbaljet: While deprecated, `rollupOptions` is still supported by Vite 8. I wonder if we should check for that here as well, like:

```
rolldownOptions: {
    input: userConfig.build?.rolldownOptions?.input 
        ?? userConfig.build?.rollupOptions?.input
        ?? resolveInput(pluginConfig, ssr)
}
```

Not 100% sure, but that might make the migration easier as Vite itself still supports it.

JonPurvis: I think this makes sense!

JonPurvis:
@juliuskiekbusch updated to Vitest 4
@pascalbaljet Added that fallback

Additionally, in `tests.yml` one of the steps is called "ESLint". I tried to rename it but don't have access to edit workflows it seems.

JonPurvis:
Thanks @pascalbaljet

sibber11:
Waiting for the merge. Thanks

phil-hudson:
pls :)

makowskid:
please merge it

pascalbaljet:
We found a potential breaking change, so we're investigating that currently. Thanks for your patience!

timacdonald:
I've replaced this one with #342, which removes any unnecessary changes top add Vite 8 support. We are still working on the breaking change.

entity:
> We found a potential breaking change, so we're investigating that currently. Thanks for your patience!

What was the breaking change? :o

timacdonald:
The behaviour of `import.meta.glob` has changed in Vite 8, which will impact Laravel apps. We are working on an upgrade path,  which is also a nicer approach for including static assets.

JonPurvis:
Thanks for taking this one over @timacdonald 🫡

entity:
> The behaviour of `import.meta.glob` has changed in Vite 8, which will impact Laravel apps. We are working on an upgrade path, which is also a nicer approach for including static assets.

Ah that's annoying! Thanks for being on it though folks

PR #342, opened 2026-03-16
[3.x] Add Vite 8 support

timacdonald:
Replaces the great work done in https://github.com/laravel/vite-plugin/pull/341 with the minimum changes required to add Vite 8 support.

We have found a breaking change we are investigating before merging this one.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -18,7 +18,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [22.x, 24.x, latest]
+        node-version: [20.x, 22.x, 24.x, latest]
 
     steps:
       - name: Checkout code
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -18,7 +18,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [20.x, 22.x, latest]
+        node-version: [22.x, 24.x, latest]
 
     steps:
       - name: Checkout code
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,5 +1,44 @@
 # Upgrade Guide
 
+## `2.x` to `3.x`
+
+- Adds support for Vite 8
+
+### Upgrade Path
+
+If you are using `import.meta.glob` to add static assets into your build, you will need to migrate them to the new `assets` config option, e.g.,
+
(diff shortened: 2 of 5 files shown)


danniejpt:
+1 please

pascalbaljet:
Up until now, we've [documented](https://laravel.com/docs/12.x/vite#blade-processing-static-assets) that you could use an unused `import.meta.glob` call to get static assets into the build manifest:

```js
import.meta.glob(['../images/**', '../fonts/**']);
```

These assets are then referenced server-side via `Vite::asset('resources/images/logo.png')`, which reads `manifest.json` at runtime.

In Vite 8, Rolldown tree-shakes the glob when the return value is unused, so the assets silently disappear from the manifest. To address this, we've added an `assets` option that uses `emitFile` in a `buildStart` hook instead, so it's not affected by tree-shaking:

```js
laravel({
    input: 'resources/js/app.js',
    assets: ['resources/images/**', 'resources/fonts/**'],
})
```

This replaces the `import.meta.glob` approach entirely. ~~It uses `fs.globSync` which requires Node >= 22, but that's fine since Node 20 is EOL in April 2026.~~

meduzen:
Ha, I’m using `vite-plugin-static-copy` for the use case you are describing.

In Vite config:

```js
import { viteStaticCopy } from 'vite-plugin-static-copy'

export default defineConfig({
  plugins: [
    laravel(),

    // Move Manifest related files at root path.
    viteStaticCopy({ targets: [
      { src: 'resources/public/manifest/*', dest: '../' }, // to `/public` root
    ] }),
```

This way I can directly reference public files in views:

```blade
<meta property="og:image" content="/android-chrome-512x512.png">
```

I guess both ways are similar.

pascalbaljet:
> Ha, I’m using `vite-plugin-static-copy` for the use case you are describing.

`vite-plugin-static-copy` does a plain file copy without hashing or versioning. The `assets` option uses `emitFile`, so files go through Vite's asset pipeline, get content-hashed filenames, and end up in the build manifest. That's what makes `Vite::asset()` work.

makowskid:
Can you please make a new release then?

juliuskiekbusch:
@makowskid this is already tagged as v3.0.0 and available via npm: https://www.npmjs.com/package/laravel-vite-plugin

PR #348, opened 2026-04-07
Add font optimization plugin

WendellAdriel:
## Summary

- Add font optimization plugin that downloads fonts from Google, Bunny, Fontsource, and local providers at build time, generating `@font-face` rules, CSS variables, utility classes, and a manifest for the framework runtime
- Support font preloading, fallback font metrics (via fontaine), per-family CSS fragments for selective loading, and unicode-range subsetting
- Dev server middleware serves cached fonts through `/__laravel_vite_plugin__/fonts/` with proper CORS and caching headers
- Production builds emit absolute asset paths in CSS so fonts resolve correctly when inlined into `<style>` tags

### Basic usage

```js
import laravel from 'laravel-vite-plugin'
import { google } from 'laravel-vite-plugin/fonts'

export default defineConfig({
  plugins: [
    laravel({
      input: 'resources/js/app.js',
      fonts: [
        google('Inter'),
      ],
    }),
  ],
})
```

### Single remote font with full API

```js
import laravel from 'laravel-vite-plugin'
import { google } from 'laravel-vite-plugin/fonts'

export default defineConfig({
  plugins: [
    laravel({
      input: 'resources/js/app.js',
      fonts: [
        google('Inter', {
          alias: 'sans',
          variable: '--font-sans',
          weights: [400, 500, 600, 700],
          styles: ['normal', 'italic'],
          subsets: ['latin', 'latin-ext'],
          display: 'swap',
          preload: [
            { weight: 400, style: 'normal' },
            { weight: 700, style: 'normal' },
          ],
          fallbacks: ['system-ui', 'sans-serif'],
          optimizedFallbacks: true,
        }),
      ],
    }),
  ],
})
```

### Advanced local font

```js
import laravel from 'laravel-vite-plugin'
import { local } from 'laravel-vite-plugin/fonts'

export default defineConfig({
  plugins: [
    laravel({
      input: 'resources/js/app.js',
      fonts: [
        // Explicit variants for full control
        local('Brand Sans', {
          alias: 'brand',
          variable: '--font-brand',
          display: 'swap',
          variants: [
            { src: 'resources/fonts/BrandSans-Regular.woff2', weight: 400 },
            { src: 'resources/fonts/BrandSans-Italic.woff2', weight: 400, style: 'italic' },
            { src: ['resources/fonts/BrandSans-Bold.woff2', 'resources/fonts/BrandSans-Bold.ttf'], weight: 700 },
          ],
          preload: [{ weight: 400, style: 'normal' }],
          fallbacks: ['system-ui', 'sans-serif'],
        }),

        // Shorthand — auto-discovers files and infers weight/style from filenames
        local('Icons', {
          src: 'resources/fonts/icons/*.woff2',
          alias: 'icons',
          preload: false,
        }),
      ],
    }),
  ],
})
```


src/fonts/fallback.ts
```diff
@@ -0,0 +1,37 @@
...
+        }
+
+        const fallbackFont = metrics.category === 'serif' ? 'Times New Roman' : 'Arial'
+
+        const { ascent, descent, lineGap, unitsPerEm } = metrics
+        if (ascent == null || descent == null || lineGap == null || unitsPerEm == null) {
+            return undefined
+        }
+
+        const sizeAdjust = 1
```
joetannenbaum, line 22: This is hardcoded to 1... what purpose does it serve?

src/fonts/providers/resolve-remote.ts
```diff
@@ -0,0 +1,109 @@
...
+import { buildResolvedFamily } from '../config.js'
+import type { FontDefinition, ResolvedFontFamily, ResolvedFontFile, ResolvedFontVariant } from '../types.js'
+
+const WOFF2_USER_AGENTS = [
+    'Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/131.0.0.0 Safari/537.36',
+    'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/131.0.0.0 Safari/537.36',
+    'Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/131.0.0.0 Safari/537.36',
+]
+
+function pickUserAgent(): string {
```
joetannenbaum, line 12: Why are we randomizing the user agent?

src/fonts/dev-server.ts
```diff
@@ -0,0 +1,84 @@
...
+
+        if (! fs.existsSync(entry.source)) {
+            res.statusCode = 404
+            res.end('Font file not found on disk')
+
+            return
+        }
+
+        const mime = FORMAT_MIME[entry.format] ?? 'application/octet-stream'
+        const data = fs.readFileSync(entry.source)
```
joetannenbaum, line 75: This blocks the event loop for every font request during development. For small woff2 files it's fine, but for OTF/TTF files that can be several MB, it'll cause noticeable lag. Should be `fs.createReadStream(entry.source).pipe(res)` or at minimum `fs.readFile` with a callback.

src/fonts/dev-server.ts
```diff
@@ -0,0 +1,84 @@
...
+
+        res.setHeader('Content-Type', mime)
+        res.setHeader('Access-Control-Allow-Origin', '*')
+        res.setHeader('Cache-Control', 'public, max-age=31536000, immutable')
```
joetannenbaum, line 79: For dev shouldn't this be `no-cache` or a very short `max-age`?

src/fonts/plugin.ts
```diff
@@ -0,0 +1,254 @@
...
+import { buildDevUrlMap, createFontMiddleware } from './dev-server.js'
+import type { Plugin, ResolvedConfig } from 'vite'
+import type { FontDefinition, ResolvedFontFamily, FallbackMetrics } from './types.js'
+
+const REMOTE_CSS_URLS: Record<string, string> = {
+    google: 'https://fonts.googleapis.com/css2',
+    bunny: 'https://fonts.bunny.net/css2',
+}
+
+const activeManifestPaths = new Set<string>()
```
joetannenbaum, line 19: This is at the module level, if someone is running multiple Vite instances this will get muddy. Claude suggests something like the following to fix:

```js
// Delete these two module-level lines:
// const activeManifestPaths = new Set<string>()
// let cleanupBound = false

// Then in configureServer, replace the activeManifestPaths/cleanupBound block with:
server.httpServer?.on('close', () => {
    try {
        if (fs.existsSync(hotManifestPath)) {
            fs.rmSync(hotManifestPath)
        }
    } catch {
        // Best-effort cleanup
    }
})
  ```

File: src/fonts/plugin.ts [hunk 6]
```diff
@@ -0,0 +1,254 @@
...
+            hotManifestPath = path.resolve(
+                path.dirname(hotFile),
+                'fonts-manifest.dev.json',
+            )
+        },
+
+        async buildStart() {
+            if (resolvedConfig.command !== 'build') {
+                return
+            }
```
joetannenbaum, line 139: Why are we building the CSS twice, once with placeholders, once with the real content? Can we just do it once in `generateBundle`?

src/fonts/css.ts
```diff
@@ -0,0 +1,149 @@
...
+    const rules: string[] = []
+
+    for (const variant of family.variants) {
+        const rangedFiles = variant.files.filter(f => f.unicodeRange)
+        const nonRangedFiles = variant.files.filter(f => ! f.unicodeRange)
+
+        for (const file of rangedFiles) {
+            const fileSrc = `url("${filePathMap.get(file.source) ?? file.source}") format("${file.format}")`
+
+            rules.push([
```
joetannenbaum, line 30: Question: The output Vite's job? It might be, but it also might be the Blade directive's job? Maybe the plugin just stores the object and the PHP side consumes it and creates the output? Food for thought.

src/fonts/css.ts
```diff
@@ -0,0 +1,149 @@
...
+                '}',
+            ].join('\n'))
+        }
+    }
+
+    return rules.join('\n\n')
+}
+
+export function generateFallbackFontFace(
+    family: string,
```
joetannenbaum, line 61: Unused?

joetannenbaum (changes requested):
Some questions/things to consider. Let me know what you think.

src/fonts/css.ts
WendellAdriel: Good question. I'd keep CSS generation on the Vite side.

The reasonL: Vite is the only place that actually knows the final asset URLs like hot-server routes in dev, hashed /build/assets/… paths in prod. Moving the logic to PHP would mean duplicating the @font-face / size-adjust / unicode-range logic and shipping an asset URL map through the manifest just so PHP can rebuild what Vite already has.

IMO, the current split feels right: Vite emits the per-family fragments + variables block, and the Blade directive just composes and filters them.

Regarding the other comment about the regex parsing in ViteFonts links with this, keeping the fragments structured (per-alias variables instead of one big :root blob) and then, PHP can concatenate directly, no CSS parsing needed.

joetannenbaum (approved):

SuperDJ:
Just a question. But could you perhaps describe the differences between this and [unplugin-fonts](https://github.com/cssninjaStudio/unplugin-fonts) with [@fontsource](https://fontsource.org/) as described in [Vuetify docs](https://vuetifyjs.com/en/getting-started/installation/#option-b-2014-install-via-unplugin-fonts-2b-40fontsource-recommended)?

WendellAdriel:
Hey @SuperDJ 

The overlap is that both approaches help generate/load `@font-face` rules from providers like Google Fonts, Fontsource, and local files.

The main difference here is the Laravel integration. This PR generates a Laravel-specific font manifest that the framework PR consumes through `@fonts` (check this PR: https://github.com/laravel/framework/pull/59584). That lets Laravel render preload tags, inline only the needed CSS, support CSP nonces, integrate with Vite hot/build modes, filter by font alias, and add font preloads to the existing Link header middleware.

So `unplugin-fonts` is a great general-purpose Vite font plugin. This is more focused on making font optimization feel native in Laravel, with the Vite plugin and framework runtime working together.

hsleewis:
@joetannenbaum Hi Joe, is there a reason why "fontaine": "^0.5.0" was added as minimum version to package.json? 0.5.0 is from 3 years ago, I understand it's the minimum, but wouldn't it be safer to just add 0.8.0 as minimum?

joetannenbaum:
@hsleewis no reason that I could see, bumped:

https://github.com/laravel/vite-plugin/pull/354

Issue #324, opened 2025-04-06
Vite triggers reload for any file in any folder


robsontenorio:
### Laravel Version

12.7.2

### PHP Version

8.4.2

### Spec

NPM 10.9.0
macOS 15.3.2

### Description

After upgrading from Laravel 11 to 12, I have noticed this strange issue where Vite triggers full page reload for any file in any folder changed. 

I am using the default settings for `vite.config.js` and `app.css`

As the docs say, not all folders will be watched.

https://laravel.com/docs/12.x/vite#blade-refreshing-on-save

### Steps To Reproduce

- Bootstrap a Laravel 12 project with some starter kit (tested with Livewire starter kit).
- Install PHP/JS dependencies.
- `npm run` dev or `yarn dev`
- Start browsing the site (without modifying any file)
- While browsing, you will notice in some pages it will refresh unexpectedly , because `storage/framework/***` changes.
- Try to modify any file like `config/***` or `database/***` , it will also refresh.

```
# Pretty random, but annoying.

1:58:51 PM [vite] (client) page reload storage/framework/views/feb9523037ff1142795d520cc7a929f4.php

```

timacdonald:
@robsontenorio, can you please check your Tailwind CSS config file and see if the `storage/framework/views` directory is included in the `content` config?

If it is there, this issue is caused by Tailwind. Tailwind creates a circular dependency which forces Vite to perform a full-page reload.

You can test this out by leaving the `refresh: true` option in the `vite.config.js` and commenting `./storage/framework/views/*.php` from the `tailwind.config.js`.

This is a known issue in Vite that I'm not sure can really be addressed: https://github.com/vitejs/vite/issues/9512#issuecomment-1208133617

robsontenorio:
Hey @timacdonald , it is a default Laravel 12 project template with Tailwind 4. There is no tailwind config file it relies on “app.css”. 

https://github.com/laravel/laravel/blob/12.x/resources/css/app.css

But, the odd part is that Vite trigger reloads for all folders. Is that a known issue ?

timacdonald:
If you comment out `@source '../../storage/framework/views/*.php';` do you still see refresh issues coming from that directory?

In regards to other files, it wasn't intentional and I'm open to excluding them by default (https://github.com/laravel/vite-plugin/issues/152#issuecomment-1937979427). This shouldn't really be the cause of your current troubles, though. Vite only refreshes when it _needs_ to even if you change files it is watching.

[...]

PR #349, opened 2026-04-17
Exclude default Laravel directories from dev server file watcher

### LINKED ISSUE #152: Vite watches the `vendor` folder
- Laravel Vite Plugin Version: 0.6.1
- Laravel Version: 9.31.0
- Node Version: 16.17.1
- NPM Version: 8.15.0
- Host operating system: Linux (Homestead, Ubuntu 18.04.5 LTS)
- Web Browser & Version: -

### Description:
Just installed Breeze (`php artisan breeze:install`) and tried to serve the assets (`npm run dev`) and I got this error, suggesting vite is watching the `vendor` folder (which I assume should be ignored, by default):

Screen Shot 2022-09-30 at 2 56 43 PM

The command `npm run build` works fine anyway.

### Steps To Reproduce:
This is my config, I tried to even defining a more restrictive `refresh` path:
```js
export default defineConfig({
    plugins: [
        laravel({
            input: [
                'resources/css/app.css',
                'resources/js/app.js',
            ],
            // none of these work
            // refresh: true,
            // refresh: false,
            refresh: [
                'resources/**',
            ],
        }),
    ],
});

```

natsuki-engr:
This PR adds a default `server.watch.ignored` configuration that excludes Laravel directories which commonly cause unnecessary full page reloads during development:

- `bootstrap`
- `database`
- `storage`
- `tests`
- `vendor`

### Motivation

Multiple users have reported that Vite triggers unexpected page reloads when files change in directories like `storage/framework/views` or `vendor` (#324, #152, laravel/laravel#6500). @timacdonald outlined a preferred approach in https://github.com/laravel/laravel/pull/6500#issuecomment-2521763613 — handle this in the plugin with sensible defaults and an escape hatch.

Resolves #152
Related: #324, laravel/laravel#6500

### Approach

- A callback-based matcher checks only the top-level directory name relative to the project root, so nested paths like `public/vendor` are not affected.
- If the user has explicitly configured `server.watch.ignored`, the plugin does not override it.
- Directories referenced in the `refresh` config (e.g. `storage/framework/views/**`) are automatically excluded from the ignore list so they remain watched.
- Only applied during `serve`, not `build`.

### Added Tests

- Ignores default Laravel directories during dev server
- Does not set watch ignored during build
- Respects user's `server.watch.ignored` config
- Keeps refresh directories watched even if they overlap ignored defaults

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -109,6 +109,21 @@ export const refreshPaths = [
     'routes/**',
 ].filter(path => fs.existsSync(path.replace(/\*\*$/, '')))
 
+/**
+ * The project-root directories excluded from Vite's dev-mode file watcher by default.
+ *
+ * Watching these directories can trigger unnecessary full page reloads and, on Linux,
+ * contributes to inotify watch exhaustion. They are matched only at the project root,
+ * so sibling directories such as `public/vendor` remain watched.
+ */
+export const watchIgnoredDirectories = [
+    'bootstrap',
+    'database',
+    'storage',
+    'tests',
+    'vendor',
+]
+
 const logger = createLogger('info', {
     prefix: '[laravel-vite-plugin]'
 })
@@ -154,6 +169,10 @@ function resolveLaravelPlugin(pluginConfig: Required<PluginConfig>): LaravelPlug
 
             ensureCommandShouldRunInEnvironment(command, env)
 
+            const watchIgnored = command === 'serve'
+                ? resolveWatchIgnored(pluginConfig, userConfig)
+                : undefined
+
             return {
                 base: userConfig.base ?? (command === 'build' ? resolveBase(pluginConfig, assetUrl) : ''),
                 publicDir: userConfig.publicDir ?? false,
@@ -189,6 +208,12 @@ function resolveLaravelPlugin(pluginConfig: Required<PluginConfig>): LaravelPlug
                             ...(userConfig.server?.hmr === true ? {} : userConfig.server?.hmr),
                         },
                         https: userConfig.server?.https ?? serverConfig.https,
+                    } : undefined),
+                    ...(watchIgnored ? {
+                        watch: {
+                            ...userConfig.server?.watch,
+                            ignored: watchIgnored,
+                        },
                     } : undefined),
                 },
                 resolve: {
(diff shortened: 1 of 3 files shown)


src/index.ts
```diff
@@ -109,6 +109,21 @@ export const refreshPaths = [
     'routes/**',
 ].filter(path => fs.existsSync(path.replace(/\*\*$/, '')))
 
+/**
+ * The project-root directories excluded from Vite's dev-mode file watcher by default.
+ *
+ * Watching these directories can trigger unnecessary full page reloads and, on Linux,
+ * contributes to inotify watch exhaustion. They are matched only at the project root,
+ * so sibling directories such as `public/vendor` remain watched.
```
timacdonald: I don't think we need this description.

```suggestion
```

src/index.ts
```diff
@@ -109,6 +109,21 @@ export const refreshPaths = [
...
+ */
+export const watchIgnoredDirectories = [
```
timacdonald: Perhaps we use `path` as I believe you could specify an path, not just a directory.

I also think this should be configurable via the project and we just this value as the default.

```suggestion
export const ignorePathsWhenWatching = [
```

src/index.ts
```diff
@@ -424,6 +449,87 @@ function resolveOutDir(config: Required<PluginConfig>, ssr: boolean): string|und
...
+ * during development but should never trigger a page reload — while leaving
+ * any directory the user has configured for refresh fully watched.
+ */
+function resolveWatchIgnored(
+    pluginConfig: Required<PluginConfig>,
+    userConfig: UserConfig,
+): ((file: string) => boolean)|undefined {
+    const userIgnored = userConfig.server?.watch?.ignored
+    if (typeof userIgnored !== 'undefined') {
+        return undefined
```
timacdonald, line 457: If the `userIgnored` has a value, should we return that value here instead of undefined? It seems like this would forget the user's settings

tests/index.test.ts
```diff
@@ -527,6 +527,61 @@ describe('laravel-vite-plugin', () => {
...
+    it("respects the user's server.watch.ignored config", () => {
+        const plugin = laravel('resources/js/app.ts')[0]
+
+        const userIgnored = ['**/my-custom-dir/**']
+        const config = plugin.config({
+            server: { watch: { ignored: userIgnored } },
+        }, { command: 'serve', mode: 'development' })
+
+        // When the user configured ignored explicitly, the plugin should not override it.
+        expect(config.server.watch?.ignored).toBeUndefined()
```
timacdonald: I don't think this test is correct. We should respect the user's values, not set it to undefined.

timacdonald:
We may also need this to be a new major version, as it is a breaking change.

timacdonald:
To test:


```
git clone https://github.com/natsuki-engr/laravel-vite-plugin.git laravel-vite-plugin
cd laravel-vite-plugin
git checkout feat/exclude-watched-directories
npm install
npm run build
npm link
```

Then in your project, install your current dependencies and then link the package in:

```
npm ci
npm link laravel-vite plugin
```

`npm run dev` should then use this branch.

timacdonald:
@natsuki-engr, I've pushed a few changes here. What do you think of this approach and does it work on your end?

natsuki-engr:
Thanks for taking this on, @timacdonald . The picomatch approach looks like a nice improvement over my directory-name matching, especially for handling paths like `storage/framework/views/**` correctly.
I'm still going through the changes in detail. I'll follow up with feedback soon, thanks for your patience!

joshmanders:
Would it make more sense to use a glob that respects gitignore? Most of these additions are gitignored stuff already, it would make sense to me that the watcher shouldn't watch files that don't matter instead of hard coding paths.

natsuki-engr:
@timacdonald Thanks for pushing those changes — tried it out and the overall approach works well on my end.

I added the missing piece from your review comment: making the ignored paths configurable via the plugin config, with `defaultIgnorePathsWhenWatching` as the default (same relationship as `refresh`/`refreshPaths`): 5fdd3b8

One thing worth calling out since it's not obvious from the diff alone: while wiring this up I noticed the path-resolving helper was also applying the `existsSync` filter to user-supplied paths, not just the defaults — meaning a path a user explicitly configured would be silently dropped if it didn't exist yet at server startup. Fixed it to trust user input as-is, matching how `refresh` never filters user-supplied paths either.

Two things I wasn't sure about while going through the changes you pushed, would appreciate your take:

**1. `userConfig.root` isn't referenced anymore**

The original implementation resolved ignored paths relative to `userConfig.root ?? process.cwd()`. The current `resolveIgnoredWatchPaths` just does `path.resolve(directory)`, which is always relative to `process.cwd()`. For a project that sets Vite's `root` to something other than the cwd, the ignored paths would be resolved against the wrong base.

This does match how `refreshPaths` (the default) already behaves, since it's also always `process.cwd()`-relative. But `refresh` has an escape hatch for custom roots via `RefreshConfig.config.root`, and there's no equivalent for `ignorePathsWhenWatching`. Was dropping `userConfig.root` intentional, or something to revisit?

**2. Is the `typeof refresh === 'boolean'` branch in `resolveRefreshMatcher` still reachable?**

```ts
function resolveRefreshMatcher(refresh: Required<PluginConfig>['refresh']): (file: string) => boolean {
    if (typeof refresh === 'boolean') {
        return picomatch([])
    }
    ...
}
```

`resolvePluginConfig` converts `refresh: true` into `[{ paths: refreshPaths }]` before it ever reaches `pluginConfig.refresh`, so as far as I can tell this branch can only ever see `false` through the public API. Not a bug, just wasn't sure if it's intentionally defensive or a leftover.

natsuki-engr:
@joshmanders Good instinct, but gitignore isn't quite the right signal here. `tests/` is a good example: test files are tracked in git, not ignored, yet editing them has zero bearing on the frontend assets Vite serves, so you still don't want them triggering a full reload. Same story for `database/` and `bootstrap/`: intentionally committed, but irrelevant to the watcher. Appreciate you raising it though, it's a fair thing to question.

PR #351, opened 2026-04-29
Add release script and publish workflow

joetannenbaum:
Adds a release.sh script that handles version bumping, tagging, and pushing from the 3.x branch, then a GitHub Actions workflow that publishes to npm with provenance whenever a release is published.


methbkts:
Any reason for using `node-version: "20"` instead of node22 or node24 ?

PR #371, opened 2026-06-11
[3.x] Use spec-compliant format() keywords for ttf, otf, and eot fonts

naveentehrpariya:
Fixes #370

The fonts plugin emits the internal format type directly into the `@font-face` `src` descriptor, producing `format("otf")`, `format("ttf")`, and `format("eot")`. These are not valid [format keywords per the CSS spec](https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face/src#font_formats), so browsers reject the source and local `ttf` / `otf` / `eot` fonts never render.

### Before

```css
src: url(".../andada-400-normal.otf") format("otf");
```

### After

```css
src: url(".../andada-400-normal.otf") format("opentype");
```

### Changes

- Added the spec `format()` keyword (`truetype`, `opentype`, `embedded-opentype`; `woff` / `woff2` unchanged) to the existing `FORMATS` config in `src/fonts/config.ts`, keeping it the single source of truth, and exposed a small `formatKeyword()` helper.
- Used the keyword in both places `src/fonts/css.ts` generates `format(...)` (regular and `unicode-range` subset rules).

The CSS parser already normalizes spec keywords on the way *in* (`FORMAT_ALIASES` in `src/fonts/css-parser.ts`); this completes the mapping on the way *out*.

Includes a test covering the emitted keyword for all five supported formats.

### DIFF
--- a/src/fonts/config.ts
+++ b/src/fonts/config.ts
@@ -5,6 +5,7 @@ import type {
     BaseFontOptions,
     FontDefinition,
     FontFormat,
+    FontFormatKeyword,
     FontProviderType,
     FontStyle,
     FontWeight,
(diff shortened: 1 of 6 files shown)


joetannenbaum:
Thanks!

PR #372, opened 2026-06-11
[3.x] Respect the resolved Vite base in generated font CSS URLs

naveentehrpariya:
Fixes #366

`local()` (and remote) fonts generate production `@font-face` URLs from `buildDirectory` alone, producing `/build/assets/...` even when the app is deployed under a subdirectory (`base: '/v2/build/'`) or served from a CDN via `ASSET_URL`. The preload links are unaffected because the font manifest stores relative file paths that the Laravel side resolves itself — but the URLs embedded in the emitted `fonts.css` and in the inline `familyStyles` 404 in those deployments.

### Before

```css
src: url("/build/assets/brand-sans-400-normal-D4x1cWHb.woff2") format("woff2");
```

### After (with `base: '/v2/build/'`)

```css
src: url("/v2/build/assets/brand-sans-400-normal-D4x1cWHb.woff2") format("woff2");
```

### Changes

- `generateBundle` now prefixes font file URLs with the resolved Vite `base` instead of hardcoding `/${buildDirectory}/`. The Laravel plugin already derives `base` from `ASSET_URL` + `buildDirectory` (and respects a user-supplied `base`), so this covers subdirectory deployments, CDN URLs, and the default case (`/build/`) identically.
- The now-unused `buildDirectory` parameter was dropped from the internal `resolveFontsPlugin()`.
- Manifest file paths remain relative — preload resolution on the Laravel side is unchanged.

Includes tests for a subdirectory base and an absolute CDN-style base, asserting both the emitted CSS and the inline `familyStyles`, and that manifest paths stay relative.

### DIFF
--- a/src/fonts/plugin.ts
+++ b/src/fonts/plugin.ts
@@ -127,7 +127,6 @@ export function assertFileRefsResolved(
 export function resolveFontsPlugin(
     fonts: FontDefinition[]|undefined,
     hotFile: string,
-    buildDirectory: string,
 ): Plugin[] {
     if (! fonts || fonts.length === 0) {
         return []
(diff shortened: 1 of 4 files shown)


joetannenbaum:
Thanks!

tinyfly:
@joetannenbaum Just a friendly nudge 🤞 to see if we can get a release with this fix?

joetannenbaum:
@tinyfly released 👍

PR #373, opened 2026-06-11
[3.x] Warn when optimized font fallbacks cannot be generated

naveentehrpariya:
Fixes #369

`optimizedFallbacks` is enabled by default, but `fontaine` is an optional peer dependency, and every failure in fallback generation is silently swallowed by a bare `catch`. As reported in #369, this leaves users with:

- No fallback `@font-face` rules and no explanation why — whether `fontaine` is missing, `readMetrics()` returns `null` for the font, or reading throws.
- `:root` variables and the manifest still referencing a `"<family> fallback"` font face that was never emitted (e.g. `--font-sans: "Vinci Sans", "Vinci Sans fallback", system-ui, sans-serif`).

### Changes

- `generateFallbackMetrics()` now accepts a `warn` callback and reports each failure mode distinctly:
  - fontaine not installed → `Optimized font fallbacks require the optional "fontaine" package. Install it, or set "optimizedFallbacks: false" on your fonts to disable the feature.`
  - metrics unreadable / incomplete → `Unable to read font metrics from [<file>]. ...`
  - read error → `Unable to generate an optimized font fallback from [<file>]: <error>`
- Warnings are routed through `this.warn()` during build and `server.config.logger.warn()` during dev, deduplicated so the fontaine hint appears once rather than per family.
- When metrics could not be generated for a family, its `optimizedFallbacks` flag is now disabled before CSS/manifest generation, so the `:root` variables and the manifest `fallbackFamily` no longer reference a font face that does not exist.

Behavior when fallback generation succeeds is unchanged.

### Tests

- Warn-callback coverage for the `null` metrics, missing-fields, and throwing `readMetrics()` paths, plus a no-warning assertion on success.
- A dedicated test simulating `fontaine` not being installed.
- An end-to-end build test asserting a warning is emitted and that neither the generated CSS variables nor the manifest reference the ungenerated fallback family.

The second report in #369 (`readMetrics` returning `null` even with fontaine installed) is a fontaine-side limitation for that specific font file — with this change it now surfaces as an explicit warning naming the file instead of failing silently.

### DIFF
--- a/src/fonts/fallback.ts
+++ b/src/fonts/fallback.ts
@@ -1,3 +1,4 @@
+import { pathToFileURL } from 'url'
 import type { FallbackCategory, FallbackEntry, FallbackMetrics } from './types.js'
 
 // Canonical OS/2 metrics for the three system fonts Fontaine uses. Values sourced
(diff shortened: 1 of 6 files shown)


joetannenbaum:
Thanks!

PR #374, opened 2026-06-11
[3.x] Honor the subsets option for google and bunny providers

naveentehrpariya:
Addresses Issue 1 of #365 (`subsets` is ignored by remote providers).

The Google and Bunny CSS2 APIs return `@font-face` rules for **every** available subset regardless of the `&subset=` parameter in the request, so the resolver downloaded and self-hosted all of them — `bunny('Inter', { subsets: ['latin'] })` still emitted all 7 subsets, and the `subsets` option silently did nothing. (Verified against both live APIs: each returns `latin`, `latin-ext`, `cyrillic`, `cyrillic-ext`, `greek`, `greek-ext`, and `vietnamese` for Inter even with `&subset=latin`.)

Both APIs label every rule with a subset comment:

```css
/* latin-ext */
@font-face {
  font-family: 'Inter';
  ...
}
```

### Changes

- The CSS parser now captures the subset label from the comment preceding each `@font-face` rule (`ParsedFontFace.subset`).
- `resolveRemoteVariants()` filters the parsed rules against the configured `subsets` before downloading any font files, so only the requested subsets are fetched and emitted.
- Rules without a subset label are kept, so unlabelled responses behave exactly as before.
- Requesting only subsets the family does not have now throws a clear error naming the available subsets instead of silently producing an empty result:
  > google returned no @font-face rules matching the requested subsets [arabic] for "Inter". Available subsets: [latin, latin-ext, ...].

### Note on default behavior

`subsets` defaults to `['latin']`, so default remote builds now self-host only the latin files instead of all subsets — this matches the documented default that previously had no effect. Since browsers already used `unicode-range` to fetch only needed subsets at runtime, this changes build output size, not rendering. Users who want more subsets opt in explicitly, e.g. `subsets: ['latin', 'latin-ext', 'cyrillic']`.

Issues 2 (variable fonts collapsing) and 3 (fontsource `unicode-range`) from #365 are separate problems and not covered here.

### DIFF
--- a/src/fonts/css-parser.ts
+++ b/src/fonts/css-parser.ts
@@ -10,16 +10,19 @@ const FORMAT_ALIASES: Record<string, FontFormat> = {
 
 export function parseFontFaceCss(css: string): ParsedFontFace[] {
     const results: ParsedFontFace[] = []
-    const ruleRegex = /@font-face\s*\{([^}]+)\}/g
+    // The Google and Bunny CSS APIs label each rule with a subset comment,
+    // e.g. `/* latin-ext */` directly above the @font-face block.
+    const ruleRegex = /(?:\/\*\s*([\w-]+)\s*\*\/\s*)?@font-face\s*\{([^}]+)\}/g
 
     let match
 
     while ((match = ruleRegex.exec(css)) !== null) {
-        const block = match[1]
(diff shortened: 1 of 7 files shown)


joetannenbaum:
Thanks!

PR #375, opened 2026-06-11
[3.x] Sanitize variable font weight ranges in emitted asset filenames

naveentehrpariya:
Addresses the generated-filename portion of #352.

When a `local()` variant uses a variable font weight range, the range is concatenated directly into the emitted asset name, producing filenames with literal spaces:

```
public/build/assets/geist-100 900-normal-CYc--oWN.woff2
public/build/assets/geist-100 900-italic-CLbs5ZTV.woff2
```

As noted in #352, spaces in asset filenames can break some build tools and CDNs, and the output is inconsistent with how Vite names the same files when they go through the `assets` pipeline.

### Changes

Whitespace in the weight is replaced with a hyphen when building the emitted asset name:

```
public/build/assets/geist-100-900-normal-CYc--oWN.woff2
```

The `@font-face` `font-weight` descriptor is unaffected and keeps the space-separated range (`font-weight: 100 900;`). Static weights (e.g. `400`) produce the same names as before.

The remaining topics in #352 (dual `src` declarations with `tech('variations')`, `local()` source entries, and `font-variation-settings`) are feature requests and not covered here.

### DIFF
--- a/src/fonts/plugin.ts
+++ b/src/fonts/plugin.ts
@@ -92,7 +92,9 @@ function emitFontAssets(
                 const source = fs.readFileSync(file.source)
                 const slug = familyToSlug(family.family)
                 const ext = file.format === 'woff2' ? '.woff2' : `.${file.format}`
-                const name = `${slug}-${variant.weight}-${variant.style}${ext}`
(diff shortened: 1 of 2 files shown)


joetannenbaum:
Thank you!

PR #379, opened 2026-06-21
Ability to specify custom certificates path

adrum:
I've been running a custom local setup to utilize Caddy/frankenphp and noticed there is not a way to specify certs from a location outside Valet/Herd. This adds a specific env var that would allow for that scenario.

I also added tests to cover the Herd config and Valet config logic.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -162,7 +162,7 @@ function resolveLaravelPlugin(pluginConfig: Required<PluginConfig>): LaravelPlug
             const env = loadEnv(mode, userConfig.envDir || process.cwd(), '')
             const assetUrl = env.ASSET_URL ?? ''
             const serverConfig = command === 'serve'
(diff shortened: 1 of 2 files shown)


timacdonald:
@adrum, can you not use the existing `VITE_DEV_SERVER_KEY` and `VITE_DEV_SERVER_CERT` environment variables?

adrum:
@timacdonald I could, but I would have to do it for every site. I like the simplicity of setting the path once in my profile and it just picks it up based on hostname, similar to Herd/Vite.

taylorotwell:
Thanks for your pull request to Laravel!

Unfortunately, I'm going to delay merging this code for now. To preserve our ability to adequately maintain the framework, we need to be very careful regarding the amount of code we include. 

If applicable, please consider releasing your code as a package so that the community can still take advantage of your contributions!

